PR #143, opened 2015-01-22
Build improvements

mwarning:
These improvements should help building ZeroTier in cross-compilation environments.


mwarning:
Looks like some other commits have been amassed here, this is not intended.

PR #150, opened 2015-02-09
Changes to build ZT for OpenWrt

mwarning:
The changes needed to build ZT for OpenWrt. Let me know what want to have changed. :-)

### DIFF
--- a/make-freebsd.mk
+++ b/make-freebsd.mk
@@ -1,5 +1,5 @@
-CC=cc
-CXX=c++
+CC?=cc
+CXX?=c++
 
 INCLUDES=
 DEFS=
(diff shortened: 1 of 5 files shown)


Red5d:
Is it possible to run ZT on DD-WRT aas well?

mwarning:
Likely, but I have never used DD-WRT.

mwarning:
The Makefile does not seem quite finished as some options from the Makefile do not seem to be set when building for OpenWrt. Stay tuned.

adamierymenko:
This will probably be merged with 1.0.3, maybe with OpenWRT becoming a supported platform if binary builds are appropriate. Is there a good way to create an OpenWRT virtual build node, such as by simulating ARM in qemu? Or maybe just pick up a cheap device?

mwarning:
OpenWrt is able to build VirtualBox/VMware images, but only for x86. It is possible to use Qemu, but I have nerver tried it. A cheap device would be a TP-LINK WR841 (4MB flash, 32MB RAM,550MHz MIPS).

maci0:
i think your best bet would be to cross compile for arm. there should be pre-built openwrt toolchains out in the wild to do that

2senel:
Works on DD-WRT, however some configuration have to be edited, especially when using jffs.

Red5d:
@2senel , what configurations need to be edited for it to work? I'm interested in getting this running on my DD-WRT router.

2senel:
I`m running it on Netgear WNDR4000 (DD-WRT v24-sp2 (04/18/14) mega - build 23919) with jffs filesystem. That means that paths in config have to be changed from /* to /jffs/*. This may vary on different routers.

Second problem is that DD-WRT does not support rc.common (as far as I know) so whole zerotier startup script in /jffs/etc/init.d has to be rewritten, it`s not so hard though.

Being lazy, I haven`t had time to rewrite that scirpt yet...So there are two things you have to do to get it run:
1. run manually zerotier-one from /jffs/usb/bin/ with &
2. use zerotier-cli as on any other machine.

If you can I can provide you precompiled .ipk for Broadcomm BCM47xx routers.

adamierymenko:
This is awesome! Definitely going to be among supported platforms.

Just so you know, this project is still very active. I've just been off coding for a short while to get some other things in order.

https://github.com/zerotier/ZeroTierOne/commit/93012b0ee529be6089f6de80705add385df574da

adamierymenko:
Oh, a quick question: are these big-endian MIPS? If so it'd be the first time this has been tested on a BE machine. Or are they 'mipsel'?

mwarning:
All ar71xx based routers are big endian (so are the WR1043v1 and wr841v9).

It can also be checked in the OpenWrt build root:
    $ grep ENDIAN .config
    CONFIG_BIG_ENDIAN=y

mwarning:
I have a better set of changes that apply to the dev branch. Time to close thi issue.

PR #151, opened 2015-02-09
minor stuff

mwarning:
A few minor issues. I suspect removing the custom swap is wrong - is it? Using size_t instead of std::size_t is only useful for some rare cases where you want to use an alternative libstc++.


mwarning:
I guess this one is outdated :>

PR #188, opened 2015-06-20
replace _networks map by vector

mwarning:
usually we use one or two networks, in these cases a vector is probably more efficient.

### DIFF
--- a/node/Node.cpp
+++ b/node/Node.cpp
@@ -311,8 +311,8 @@ ZT1_ResultCode Node::processBackgroundTasks(uint64_t now,volatile uint64_t *next
 ZT1_ResultCode Node::join(uint64_t nwid)
 {
 	Mutex::Lock _l(_networks_m);
-	std::vector< SharedPtr<Network> >::const_iterator nwi = std::lower_bound(_networks.begin(), _networks.end(), nwid, NetworkComparator());
-	if(nwi == _networks.end() || (*nwi)->id() != nwid) {
+	SharedPtr<Network> nw = _network(nwid);
+	if(!nw) {
 		_networks.push_back(SharedPtr<Network>(new Network(RR,nwid)));
 		std::sort(_networks.begin(),_networks.end());
 	}
@@ -412,10 +412,10 @@ ZT1_PeerList *Node::peers() const
 ZT1_VirtualNetworkConfig *Node::networkConfig(uint64_t nwid) const
 {
 	Mutex::Lock _l(_networks_m);
-	std::vector< SharedPtr<Network> >::const_iterator nwi = std::lower_bound(_networks.begin(), _networks.end(), nwid, NetworkComparator());
-	if(nwi != _networks.end() && (*nwi)->id() == nwid) {
+	SharedPtr<Network> nw = _network(nwid);
+	if(nw) {
 		ZT1_VirtualNetworkConfig *nc = (ZT1_VirtualNetworkConfig *)::malloc(sizeof(ZT1_VirtualNetworkConfig));
-		(*nwi)->externalConfig(nc);
+		nw->externalConfig(nc);
 		return nc;
 	}
 	return (ZT1_VirtualNetworkConfig *)0;
--- a/node/Node.cpp
+++ b/node/Node.cpp
@@ -236,13 +236,14 @@ ZT1_ResultCode Node::processBackgroundTasks(uint64_t now,volatile uint64_t *next
 			std::vector< SharedPtr<Network> > needConfig;
 			{
 				Mutex::Lock _l(_networks_m);
-				for(std::map< uint64_t,SharedPtr<Network> >::const_iterator n(_networks.begin());n!=_networks.end();++n) {
-					SharedPtr<NetworkConfig> nc(n->second->config2());
-					if (((now - n->second->lastConfigUpdate()) >= ZT_NETWORK_AUTOCONF_DELAY)||(!nc))
-						needConfig.push_back(n->second);
+				for(std::vector< SharedPtr<Network> >::const_iterator n(_networks.begin());n!=_networks.end();++n) {
+					SharedPtr<NetworkConfig> nc((*n)->config2());
+					if (((now - (*n)->lastConfigUpdate()) >= ZT_NETWORK_AUTOCONF_DELAY)||(!nc))
+						needConfig.push_back(*n);
 					if (nc)
 						networkRelays.insert(networkRelays.end(),nc->relays().begin(),nc->relays().end());
 				}
+				std::sort(_networks.begin(),_networks.end());
 			}
 
 			// Request updated configuration for networks that need it
@@ -310,19 +311,24 @@ ZT1_ResultCode Node::processBackgroundTasks(uint64_t now,volatile uint64_t *next
 ZT1_ResultCode Node::join(uint64_t nwid)
 {
 	Mutex::Lock _l(_networks_m);
-	SharedPtr<Network> &nwe = _networks[nwid];
-	if (!nwe)
-		nwe = SharedPtr<Network>(new Network(RR,nwid));
+	std::vector< SharedPtr<Network> >::const_iterator nwi = std::lower_bound(_networks.begin(), _networks.end(), nwid, NetworkComparator());
+	if(nwi == _networks.end() || (*nwi)->id() != nwid) {
+		_networks.push_back(SharedPtr<Network>(new Network(RR,nwid)));
+		std::sort(_networks.begin(),_networks.end());
+	}
 	return ZT1_RESULT_OK;
 }
 
 ZT1_ResultCode Node::leave(uint64_t nwid)
 {
 	Mutex::Lock _l(_networks_m);
-	std::map< uint64_t,SharedPtr<Network> >::iterator nw(_networks.find(nwid));
-	if (nw != _networks.end()) {
-		nw->second->destroy();
-		_networks.erase(nw);
+	std::vector< SharedPtr<Network> >::iterator nwi = std::lower_bound(_networks.begin(), _networks.end(), nwid, NetworkComparator());
+	if(nwi != _networks.end() && (*nwi)->id() == nwid) {
+		(*nwi)->destroy();
+		// erase element (replace by last)
+		*nwi = _networks.back();
+		_networks.pop_back();
+		std::sort(_networks.begin(),_networks.end());
 	}
 	return ZT1_RESULT_OK;
 }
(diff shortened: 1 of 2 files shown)


janjaapbos:
I actually use a lot of networks, (20+ on my laptop) and I think this will increase further. ;-)

mwarning:
ah, alright, then we may need to add a sort&binary_search - or just leave it be. :)

mwarning:
Some thoughts.. Memory consumption does not matter here. What matters is access speed. Since uClibc++ map is a list, doing a binary search on a vector is likely to be much faster. This approach might also be faster compared to stdc++ map. As long the vector is not really big. I will see if I can write it with binary search on a vector.

keesbos:
Good. As I understand *network(uint64_t nwid)* is called for every packet, so speed is important.

To my knowledge, hashes are better (fatser) for large data sets. However, the dataset we will have is probably small (<1000) and we have integers to compare in the binary search. So,  after all a binary search might generally be faster that 'create hash and traverse linked list' or whatever algorithm the map may use.

Maybe its possible to create a tiny test program and do a test to time/compare the two approaches?
(for say, 10, 100, 1000, 10000 nwids).

mwarning:
I wrote a simple test program (outside of ZT...). For 10 elements, vector is 50% faster. Speed is on par for ~80 elements. For 10000 elements, map is 40% faster. For 100000 it is on par again. It is hard to measure...

keesbos:
I guess one would only have lots of networks on fast cpus...

janjaapbos:
Yes, would be good to aim for the best performance at the lowest end, where the number of active networks is probably low too.

keesbos:
BTW. This might be the reason that you get varying mileage:
http://stackoverflow.com/questions/454762/vector-or-map-which-one-to-use

In short: map uses also binary sort (as opposed to std::tr1::unordered_map). Difference map and vector is more or less the way memory is allocated.

keesbos:
I really couldn't resist diving into binary / hash comparison ;-)

In short: hash lookup is pretty constant, binary search depends on haystack size.

```c++
#include <iostream>
#include <unordered_map>
#include <map>
#include <ctime>


volatile int acc = 0;

void testMap(int size, long int count) {
        std::clock_t start;
        double duration;
        std::map< int, int > m;
        //std::unordered_map< int, int > m;

        for( int i=0; i < size; i++ )
                m[i] = i; 

        start = std::clock();

        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ )
                        acc = m.find(i)->first;
        }

        duration = ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
        printf("%d: %f\n", size, duration);
}

int main(int argc,char **argv) {

        testMap( 2, 50000000L);
        testMap(10, 10000000L);
        testMap(100, 1000000L);
        testMap(1000, 100000L);
        testMap(10000, 10000L);
        return 0;
}
```
Output for map:
```
2: 6.236409
10: 8.630706
100: 17.838460
1000: 21.759229
10000: 25.675577
```
Output for unordered_map:
```
2: 8.974842
10: 8.991932
100: 8.855265
1000: 9.092861
10000: 9.040554
```
However, unordered_map is from c++11

I guess the best would be something that stores/uses binary tree for small sizes (<10) and stores/uses hashes for larger collections.

mwarning:
:-) Anyway, the fastest approach in this case would probably a closed hash table: index = nwid % vector_size; lookup object on index, if it does not match, try index+1 until the object is found.

keesbos:
I wouldn't be surprised when (for small sizes) an ordinary array would
outperform c++ structures.

keesbos:
OK. Just tried an ordinary C array (binary lookup).

First the results:
```
size: map, unordered_map, vector, array
2 : 6.190372, 8.972926, 7.816989, 0.721049
10 : 8.786468, 9.003411, 10.743592, 1.733225
100 : 17.799845, 8.945659, 15.980356, 4.855754
1000 : 21.577814, 8.950351, 23.042304, 10.362637
10000 : 25.912142, 8.983744, 27.515797, 14.062496
```
Smaller numbers is better.

If we _really_ want to make some difference, I would go for a hashed map in plain C array (not in test :-) ).

Now the test code. It's dirty, just to get some idea of the differences. Probably quite easy to optimize etc. My C code experience is somewhat rusty (from centuries ago) and my C++ experience just a few days ;-)

It's just meant to test lookups, since that's almost the only thing we do with Network.

```c++
#include <iostream>
#include <map>
#include <unordered_map>
#include <vector>
#include <ctime>


class Network
{
public:
        inline uint64_t id() const throw() { return _id; }
        Network(uint64_t id) {
                _id = id;
        }
private:
        uint64_t _id;
};

volatile int acc = 0;

double testMap(int size, long int count) {
        std::clock_t start;
        std::map< int, Network* > m;

        for( uint64_t i=0; i < size; i++ )
                m[i] = new Network(i);

        start = std::clock();
        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ ) {
                        acc = m.find(i)->first;
                }
        }
        return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}

double testUnorderedMap(int size, long int count) {
        std::clock_t start;
        std::unordered_map< int, Network* > m;

        for( uint64_t i=0; i < size; i++ )
                m[i] = new Network(i);

        start = std::clock();
        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ ) {
                        acc = m.find(i)->first;
                }
        }
        return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}

double testVector(int size, long int count) {
        std::clock_t start;
        std::vector< Network* > m(size);

        for( uint64_t i=0; i < size; i++ )
                m[i] = new Network(i);

        start = std::clock();
        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ ) {
                        std::vector< Network* >::iterator iter = std::lower_bound(m.begin(), m.end(), m[i]);
                        if (iter != m.end()) {
                                acc = i;
                                //acc = (*iter)->id();
                        }
                }
        }
        return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}

double testArray(int size, long int count) {
        std::clock_t start;
        Network *m[10000];

        for( int i=0; i < size; i++ )
                m[i] = new Network(i);

        start = std::clock();
        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ ) {
                        int j = size/2, lower=0, upper=size-1, j0 = -1, j1;
                        while (i != j) {
                                if (m[j]->id() < i)
                                        lower = j;
                                else
                                        upper = j;
                                j1 = lower + (upper - lower) / 2;
                                if (j1 == j) {
                                        if (m[j]->id() < i)
                                                j1++;
                                        else
                                                j1--;
                                }
                                j = j1;
                                if (j == j0) {
                                        // not found
                                        j0 = -1;
                                        break;
                                }
                                j0 = j;
                        }
                        if (j1 >= 0) {
                                // printf("found %d at %d\n", i, j);
                                acc = j;
                        }
                }
        }
        return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}

void test(int size, long int count) {
        printf("%d : ", size);
        printf("%f, ", testMap(size, count));
        printf("%f, ", testUnorderedMap(size, count));
        printf("%f, ", testVector(size, count));
        printf("%f", testArray(size, count));
        printf("\n");
}

int main(int argc,char **argv) {

        std::cout << "START\n";
        printf("size: map, unordered_map, vector, array\n");
        test( 2, 50000000L);
        test(10, 10000000L);
        test(100, 1000000L);
        test(1000, 100000L);
        test(10000, 10000L);
        std::cout << "DONE\n";
        return 0;
}
```

mwarning:
This is simple closed hash map.

```
double testHashArray(int size, long int count) {
	std::clock_t start;
#define SIZE 10000
	Network *m[SIZE];

	//setup array
	memset(m, 0, sizeof(m));
	for( int i=0; i < size; i++ ) {
		int idx = i % SIZE;
		bool ok = false;
		for(int x = 0; x < SIZE; ++x, ++idx) {
			int p = idx % SIZE;
			if(m[p] == NULL) {
				m[p] = new Network(i);
				ok = true;
				break;
			}
		}
		if(!ok) {
			//not space left in array - need to resize
			exit(1);
		}
	}

	start = std::clock();
	for( int n=0; n < count; n++ ) {
		for( int i=0; i < size; i++ ) {
			int idx = i % SIZE; //start looking at this position
			for(int x = 0; x < SIZE; ++x, ++idx) {
				//int p = idx % SIZE;
				int p = (idx >= SIZE) ? (idx - SIZE) : idx;
				if(m[p] == NULL) {
					//not found
					break;
				}

				if(m[p]->id() == i) {
					//found
					acc = i;
					break;
				}
			}
			//not found
		}
	}
	return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}
```

keesbos:
Cool. I've rerun the test with the hash array included.

```
size: map, unordered_map, vector, bin_array, hash_array
2 : 6.152070, 8.899776, 7.626522, 0.705462, 0.848976
10 : 8.741328, 9.042351, 10.250206, 1.747433, 0.802372
100 : 17.743314, 8.871051, 15.663290, 4.440715, 0.797141
1000 : 21.243506, 8.929548, 22.553460, 10.128337, 0.787725
10000 : 25.021659, 8.970584, 26.639330, 14.045490, 0.793292
```

However, I think the timings for this hash array is a bit misleading. There is collision handling, but in the test case there are no collisions generated. We'll need to handle that too for a good comparison. I've added some collisions in the test. Since this extra code makes the test slower anyway, I've used it without actual creating collisions and with (`k = i & 1 ? i + size + 0 : i;` versus `k = i & 1 ? i + size + 1 : i;`)

```
size: no collisions, half collisions
2 : 1.145949, 1.577047
10 : 1.110155, 1.532960
100 : 1.127447, 1.545426
1000 : 1.119299, 1.509364
10000 : 1.115421, 1.505513
```
So, the effect of having half of the ids colliding in the 'hash' is not too bad :-)

If Adam thinks it's worthwhile to continue with this, we should probably create a class without loosing performance somehow and without possible overflow etc.


Altered `testHashArray`:
```c++
double testHashArray(int size, long int count) {
        std::clock_t start;
#define SIZE size
        Network **m;
        int k;

        //setup array
        m = (Network **) malloc(size * sizeof(m));
        memset(m, 0, sizeof(m));
        for( int i=0; i < size; i++ ) {
                // create collision
                k = i & 1 ? i + size + 1 : i;

                int idx = k % SIZE;
                bool ok = false;
                for(int x = 0; x < SIZE; ++x, ++idx) {
                        int p = idx % SIZE;
                        if(m[p] == NULL) {
                                m[p] = new Network(k);
                                ok = true;
                                break;
                        }
                }
                if(!ok) {
                        //no space left in array - need to resize
                        exit(1);
                }
        }

        start = std::clock();
        for( int n=0; n < count; n++ ) {
                for( int i=0; i < size; i++ ) {
                        // create collision
                        k = i & 1 ? i + SIZE + 1 : i;

                        int idx = k % SIZE; //start looking at this position
                        for(int x = 0; x < SIZE; ++x, ++idx) {
                                //int p = idx % SIZE;
                                int p = (idx >= SIZE) ? (idx - SIZE) : idx;
                                if(m[p] == NULL) {
                                        //not found
                                        break;
                                }

                                if(m[p]->id() == k) {
                                        //found
                                        acc = k;
                                        break;
                                }
                        }
                        //not found
                }
        }
        return ( std::clock() - start ) / (double) CLOCKS_PER_SEC;
}
```

adamierymenko:
Performance-wise, you'd probably have to be a member of over a hundred networks before a vector would be slower than a map -- as some of your tests show.

Maps were used a lot in the ZeroTier code because it's a simple straightforwardly "correct" representation, but it's not necessarily the most efficient abstraction. As we optimize we'll probably be factoring them out in a bunch of places.

The biggest caveat is that we must take care to check whether any of the code depends on the ordering of the data structure. Maps are trees with implicit ordering by key, so replacing them with an unordered structure should occur with care. If insert / erase operations are rare, sorting a vector is pretty cheap... but if those operations are common then it might be faster to use the tree-based map that implicitly sorts.

adamierymenko:
Closing since I merged it manually into -dev -- master is for releases.

mwarning:
oh, missed that it was aimed at the master branch. :=)

PR #193, opened 2015-06-23
replace _rxQueue list by vector for faster memory access and less all…

mwarning:
…ocations


mwarning:
_rxQueue does not seem to be used as an ordered container. But some more insight might be required.

adamierymenko:
The typical length of this structure is short, but I was concerned about the use of a vector due to wasted memory. If you suddenly get a ton of packets, it might get kind of long (maybe 10-20 packets) for a little while, then back to zero... but vector won't automatically reclaim.

adamierymenko:
the multi map is kind of unnecessary though... thing thing is the entries are large (full Packet plus fragments buffer)

adamierymenko:
Probably won't merge this one... closing for now. (See our IRC discussion.)

PR #195, opened 2015-06-24
Fix signing verification

keesbos:
From my understanding, this is the problem that's fixed:

The only signing that's accepted is what we (the controller) signed.

	if (cert.signedBy() != controller()) then fail

So, since we're only accepting certs that should be singed by ourself (i.e. the controller), there's no need to get a Peer entry for ourself (which will fail). Just use our own (the controller) identity to verify the signing.

### DIFF
--- a/controller/SqliteNetworkController.cpp
+++ b/controller/SqliteNetworkController.cpp
@@ -1176,7 +1176,14 @@ unsigned int SqliteNetworkController::_doCPGet(
 						sqlite3_bind_text(_sGetIpAssignmentsForNode2,2,addrs,10,SQLITE_STATIC);
 						bool firstIp = true;
 						while (sqlite3_step(_sGetIpAssignmentsForNode2) == SQLITE_ROW) {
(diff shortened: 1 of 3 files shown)


keesbos:
Hmm, I'm not quite sure this is the solution to the problem I'm seeing....

keesbos:
OK. I the first commit fixes one of the problem(s) I saw, but breaks other signing. Fixed that by reverting the deletion of code in the first commit and leaving the new code in for 'self signed' certs.

Furthermore, fixed lookup of networks. The lookup missed networks that where available and in the table.

keesbos:
I don't think iter will ever be _networks.end():
```c++
for(iter(_networks.begin()) ; iter!=_networks.end() ; ++iter)
```
I guess your suggestion of being unsorted is probably correct. We'll have to find that before reverting to std::lower_bound.

keesbos:
Added also a fix for the controller that reported faulty ipv4 assignments

PR #196, opened 2015-06-26
Fixes for controller

keesbos:
Reordered commits for pull #195 

Now just one commit for fixing cert checking and one for reporting back ipv4 assignments of for mebers.


adamierymenko:
So this should supersede 195?

keesbos:
Yep. And rebased :-)

PR #197, opened 2015-06-26
Default ethertypes

keesbos:
When there are no rules defined, nothing is permitted. In principle a correct implementation, but a zt-net without traffic doesn't make sense. So, I propose to allow IPv4, ARP, and IPv6 ethernet frames when no rules are defined. I think that's kind of the 'standard' use case.


keesbos:
Should go to dev :-(

adamierymenko:
Hrm... problem is this conflicts with upcoming changes. The etherTypes thing is getting replaced by a full-blown OpenFlow-like rules engine.

adamierymenko:
In a proper rules engine 'do nothing' is and should be the default behavior if you fall off the end of the rules queue. I think if there's a problem here it should be fixed at the UI level -- the UI should create new networks with sane defaults. We *could* make the default settings for a new network in the NetworkController do this.

adamierymenko:
In other words, make this the default rule set if no rules are defined when a new network is created.

PR #198, opened 2015-06-26
Set default accepted ethertypes if no rules are defined

keesbos:
When there are no rules defined, nothing is permitted. In principle a correct implementation, but a zt-net without traffic doesn't make sense. So, I propose to allow IPv4, ARP, and IPv6 ethernet frames when no rules are defined. I think that's kind of the 'standard' use case.

### DIFF
--- a/controller/SqliteNetworkController.cpp
+++ b/controller/SqliteNetworkController.cpp
@@ -50,6 +50,7 @@
 #include "../node/InetAddress.hpp"
 #include "../node/MAC.hpp"
 #include "../node/Address.hpp"
+#include "../node/Switch.hpp"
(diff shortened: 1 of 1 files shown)


adamierymenko:
See my replies to the other one you just closed. :)

keesbos:
:-) Yep. I'll close this one too

adamierymenko:
Yeah... I think that's the wrong place to put it. It should be (if anywhere) in new network creation. If they create a new network with no rules, it could set those as default rules.

adamierymenko:
I agree that 99% of users will want IPv4, ARP, and IPv6 enabled on a normal 'flat' network. But our future rules engine is going to support all kinds of weird and creative things. :)

PR #199, opened 2015-06-27
Fix for use of std::unique on std::vector

keesbos:
The std::unique function does update the values in the vector, but does not alter the length (i.e. the return value of .end() )

Example:
With the rules I had ethertypes 0000, 0800, 0000. The config (saved in /var/lib/zerotier/networks.d) had 0000, 0800, 0800. After popping the elements where end != vector.end(), the values became 0000, 0800.

### DIFF
--- a/controller/SqliteNetworkController.cpp
+++ b/controller/SqliteNetworkController.cpp
@@ -389,9 +389,9 @@ NetworkController::ResultCode SqliteNetworkController::doNetworkConfigRequest(co
 					allowedEtherTypes.push_back(et);
 			}
 			std::sort(allowedEtherTypes.begin(),allowedEtherTypes.end());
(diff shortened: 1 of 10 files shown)


adamierymenko:
You've got to be kidding me. I'm [astonished](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), therefore this is IMO a bug in STL.

I think I might fix this myself in a cleaner way by creating Utils::vectorUnique(). Sigh.

adamierymenko:
Oh right! It never does! Duh!

It can't. It gets iterators, not the object in question. I'll make a version that does what I want.

adamierymenko:
The way to do this is:

    container.erase(std::unique(container.begin(),container.end()),container.end());

keesbos:
OK. Do you want me to do that, or do-you-do-it-yourself ?

adamierymenko:
Doing it now.

keesbos:
:-)

PR #200, opened 2015-06-27
Make rules either general or member specific

keesbos:
When the nodeId is NULL in the Rule table, the rule is generic and used on member nodes that do not have specific rules set.

This way, it's possible to have general rules for most of the members/nodes, while specific members can use/accept other rule sets (e.g. for bridging and vrrp etc.)


adamierymenko:
I think maybe nodeId should be redefined. Right now it's confusing-- in most tables nodeId selects which node this applies to, but in Rule it's a field in the rule table.

Instead, I propose creating a portId field which is the nodeId of the port on the virtual switch. NodeID can then behave as you suggest -- NULL for network-wide rules or nodeId for a specific node that gets this rule.

adamierymenko:
I'm doing some changes to the DB now, so I will do this after my current route/IP assignment range changes.

PR #259, opened 2015-11-30
Increase maximum number of paths per IP scope to 4

janjaapbos:
Otherwise, local discovered routes are ignored.
Don't know what the best value would be. Taking 4 for now.


adamierymenko:
Yes, I think this is okay. The limit is there to prevent flooding so I'll review from that POV but increasing it a little won't hurt anything.

PR #318, opened 2016-03-28
Fix escaping in controller/README.md

hexchain:


hexchain:
My bad for not reading README about contributions. Will reopen against `edge`.

PR #335, opened 2016-06-07
.deb installer missing Section header

nucleardreamer:
Hi guys! 

The `Section` header on the deb installer `control` file is definitely optional, however, it breaks some automation using local deb repos. An example is installing with [reprepro](http://mirrorer.alioth.debian.org/), you will just get:
```
No section given for 'zerotier-one', skipping.
```  
This isn't particularly a problem because there is a flag to ignore this, but it breaks when doing auto-installation and sticking to the "rules". It would also be good in advance of a possible PPA or mirror opportunities ;)

### DIFF
--- a/ext/installfiles/linux/DEBIAN/control.in
+++ b/ext/installfiles/linux/DEBIAN/control.in
@@ -4,6 +4,7 @@ Maintainer: ZeroTier, Inc. <contact@zerotier.com>
 Priority: optional
 Version: __VERSION__
 Installed-Size: 1024
+Section: net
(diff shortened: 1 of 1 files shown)


adamierymenko:
Thanks but we're totally re-doing packaging right now from scratch for both DEB and RPM with the hope of getting this into core distros. The existing packages are very bare bones and not quite up to distro specs.

adamierymenko:
We will make sure this is fixed in the next rev of packaging.

nucleardreamer:
Sounds good to me. Feel free to @ me on any threads if you guys are looking for help!

PR #401, opened 2016-10-16
Run as nonroot user on Linux (with CAP_NET_ADMIN and CAP_NET_RAW added).

zielmicha:
- ZT will only drop root privileges if zerotier-one user exists. It is created by Debian postinst script - in other cases the user has to be created by administrator.
- Linux >=4.3 with ambient capabilities is required, otherwise ZT will silently run as root
- "-U" option now also disables privileges dropping

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,0 +1,9 @@
+#!/bin/sh -e
+
+case "$1" in
+  configure)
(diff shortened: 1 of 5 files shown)


adamierymenko:
Merged! Will test and review some more but this looks like really great high-quality work. Thanks!

adamierymenko:
This addresses issue #401 -- which will be closed soon.

PR #428, opened 2016-12-27
NetBSD support

danielskowronski:
I've added (very) experimental support for NetBSD - highly portable BSD distribution.

Only osdep/*EthernetTap needed to be changed (plus some adding \_\_NetBSD\_\_ in some places in code and custom makefile). As on older OS X getifmaddrs() and related are missing in NetBSD kernel so I've added them from FreeBSD. Implementation of creating tap device is quick and dirty but for now - it works.


danielskowronski:
Tested with `NetBSD 7.0.2 (GENERIC.201610210724Z) i386` and `gcc version 4.8.4 (nb2 20150115)`

osdep/NetBSDEthernetTap.cpp
```diff
@@ -0,0 +1,484 @@
...
+				int exitcode = -1;
+				::waitpid(cpid,&exitcode,0);
+			} else throw std::runtime_error("fork() failed");
+
+			cpid = (long)vfork();
+			if (cpid == 0) {
+				string tmp;
+				sprintf((char*)tmp.c_str(), "%d", i);
+				string minor = tmp.c_str();
+				::execl("/sbin/mknod","/sbin/mknod",devpath,"c","169",minor.c_str(),(const char *)0);
```
krytarowski: `cd /dev && ./MAKEDEV tap`

osdep/NetBSDEthernetTap.cpp
```diff
@@ -0,0 +1,484 @@
...
+	} else if (cpid > 0) {
+		int exitcode = -1;
+		::waitpid(cpid,&exitcode,0);
+	}
+
+	cpid = (long)vfork();
+	if (cpid == 0) {
+		string tmp="/dev/";
+		tmp+=_dev.c_str();
+		::execl("/bin/rm","/bin/rm",tmp.c_str(),(const char *)0);
```
krytarowski: `unlink(2)`?

Makefile
```diff
@@ -16,3 +16,7 @@ endif
 ifeq ($(OSTYPE),OpenBSD)
 	include make-freebsd.mk
 endif
+
+ifeq ($(OSTYPE),NetBSD)
+	include make-netbsd.mk
```
krytarowski: Is it possible to unify `make-freebsd.mk` and `make-netbsd.mk` into `make-bsd.mk`?

adamierymenko:
Hmm... will have to see if this works against dev. Thanks!

adamierymenko:
Now that OpenBSD is in `dev` I think we will have to look at this. Will have to be ported to `dev`.

Makefile
adamierymenko: In `dev` we now have a make-bsd which should do this.

adamierymenko:
We will have to make a NetBSD CI VM on our VM host too.

krytarowski:
I'm looking forward to it.

PR #449, opened 2017-03-13
Add make command for linux-build-farm

makersphereLabs:
Without the `make one` command in the Makefile, building the .deb packages using the linux-build-farm (which runs `make debian` inside a Docker container) fails when executed on a **non-Linux machine**. (e.g. macOS)


adamierymenko:
Closing since linux-build-farm is deprecated. We will be publishing a torrent or something of our new qemu-chroot based multi-arch system.

PR #450, opened 2017-03-13
Wait for ZT networks to become available

makersphereLabs:
It's not very pretty **but it works**. 😅
If you're running services like SSH, Docker etc. which _bind to a specific ZeroTier interface_, they will most likely fail during startup because `systemd` doesn't waits for ZT networks to come up.
Adding a this "dynamic" delay doesn't impact startup time in any substantial way.
Without this _small hack_, it's just random luck when e.g. Docker successfully starts after all configured ZT (networks) interfaces are up.

Tested on **Ubuntu 16.04** with ZeroTier One v1.1.17 and Docker 17.03 CE

### DIFF
--- a/debian/zerotier-one.service
+++ b/debian/zerotier-one.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=ZeroTier One
+Documentation=https://www.zerotier.com/manual.shtml
 After=network.target
 Before=ssh.service docker.service
 
--- a/debian/zerotier-one.service
+++ b/debian/zerotier-one.service
@@ -1,9 +1,15 @@
 [Unit]
 Description=ZeroTier One
 After=network.target
+Before=ssh.service docker.service
 
 [Service]
 ExecStart=/usr/sbin/zerotier-one
+ExecStartPost=/bin/bash -c '\
+until [ "`find /var/lib/zerotier-one/networks.d -type f -name "????????????????.conf" | \
+wc -l`" == "`find /sys/devices/virtual/net -type d -name "zt*" | wc -l`" ]; do \
+sleep 0.5; \
+done; echo "ZeroTier One networks are online";'
 Restart=always
 KillMode=process
 


misuzu:
I think there is a better approach:
https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/connman-wait-online.service.in

makersphereLabs:
@misuzu Well, this **doesn't work** with ZeroTier... Already tried that.
Because ZT networks are _virtual interfaces_, it's a bit more complicated.
In a nutshell, `systemd` starts `zerotier-one.service` and moves on with the startup sequence, but the ZT interfaces (`zt0, zt1, etc.`) need some time to become available. Just using `After=` or `Before=` doesn't do the trick because `systemd` does not wait/know about these virtual interfaces.

keesbos:
I'm not sure if this will be generally favorable. It assumes you want all interfaces to be up before continuing (if I read it right). Per interface could be done with normal if-scripts. E.g. something like:

/etc/network/interfaces:
```
iface zt0 inet static
	address 192.168.0.10
	netmask 255.255.255.0
	broadcast 192.168.0.255
```

/etc/network/if-pre-up.d/zt0:
```
#!/bin/sh
# Start zerotier-one (if it's not started yet) and wait for zt0
/etc/init.d/zerotier-one start
echo -en "Waiting for zt0"
while ! ifconfig zt0 >/dev/null 2>&1; do echo -en '.' ; sleep 1 ; done
echo
```

makersphereLabs:
@keesbos Na, not really. Maybe the PR desc. is a bit misleading. 😬
It's not about the physical interfaces `eth0, eth1`... For those ifaces _if-pre-up scripts_ work just fine.
But ZeroTier networks are **virtual interfaces** and their IPs are assigned via ZT.
Also messing with the `/etc/network/interfaces` config doesn't work here and isn't really scalable, when you need to provision multiple networks & machines. Btw. already tried that too. 😅

Since the last few releases, Ubuntu uses `systemd` to start services, incl. Docker, ZeroTier, SSH, etc.
Which means that you have to modify the `.service` file, to make sure the ZT networks are up before continuing the startup sequence. Most people expect ZeroTier interfaces to behave like "normal", physical network connections. (They're obviously not.) This tweak ensures that these _virtual ifaces_ are created & available before you launch other services. Now they can be used like any other (physical) interface. ✌️

**Use Case:** E.g. SSH binds to zt0, for remote access only via a ZT network.

victorqhong:
@makersphereLabs Try changing `After=network.target` in the unit file to `After=network-pre.target`. That should make the service start a bit earlier and fixed it for me.

makersphereLabs:
@victorqhong Interesting, **which OS** are you using and **how many ZT networks** are configured?
Have you tried to verify the startup order with something like `systemd-analyze plot > plot.svg` to make sure it doesn't come down to _random luck_ if it works or not? Especially important when you're using SSH only via ZeroTier. 🤔

makersphereLabs:
We could add some kind of **timeout** function, to make sure the script doesn't block the boot process.
This would prevent a scenario, like @zielmicha described it in a [comment](https://github.com/zerotier/ZeroTierOne/issues/451#issuecomment-286886173) to #451.

@adamierymenko What do you think?
And is this tweak _too specific to Ubuntu_? Should we add version for other _systemd-based OS_?
We would need to update the `zerotier-one.spec` tho...

lloesche:
Why not change it to be a built-in that works on all platforms. Maybe add a `zerotier-wait` personality that blocks until a given network or all networks are connected.

Then you could have something like `ExecStartPost=-/usr/sbin/zerotier-wait all` in your service unit. Folks who want to block until the networks are up create the symlink, others don't.

Or if you don't like user created symlinks in /usr/sbin/ maybe allow for an optional `EnvironmentFile=-/etc/zerotier/wait` with `ZT_WAIT_FOR=all` in it and process that environment variable in the tool itself with the default (`ZT_WAIT_FOR` is empty/undefined) being to not wait and return immediately.

Just some ideas to make this otherwise really useful feature more portable.

dch:
as a non-systemd os user, I've found that I can't start firewall successfully until zt interfaces are up, which is awkward for automated startup. Something that supports this would be useful, but I may be able to get away with a simpler solution on FreeBSD at least. Still testing.

dch:
FWIW I've found a similar constraint - while I can ensure via rcorder(8) that network interfaces are created prior to zerotier, and that zerotier starts before the firewalls do, but the firewalls themselves *still* need assigned IPs before "things just work". Making zerotier not return from startup until all interfaces are up and have an IP assigned would be great, ideally via option flag. `--wait-for-IPs` sort of thing.

fschulze:
This might be the reason why my firewall port forwards on freebsd sometimes don't work. Is there a way to do stuff when a zero tear network comes up? Then I could reload the firewall config. Waiting on zerotier IPs to become available is risky if other firewall rules are required to get access to a server at all.

dch:
@fschulze the changes I made are as follows:

- make zt dependent on ipfw & pf startup
- tweak my `pf.conf` slightly:

normally you'd use this approach `zero_net = $zero_if:network` and then pf works out the correct ip range from the live interface. If you use instead `zero_net = "10.144.0.0/16"` then pf doesn't have to wait for actual IP addresses to be present on the tap interface, and startup + rules can proceed without issue.

the changes for zt in ports to support this will go in this weekend, but you can try it out with the following manual change to `/usr/local/etc/rc.d/zerotier`

```diff
6c6,7
< # REQUIRE: LOGIN
---
> # REQUIRE: FILESYSTEMS netif
> # BEFORE:  pf ipfw
```

fschulze:
@dch that change doesn't seem to be in ports yet, do you have a link to the issue/PR for it?

dch:
my bad, had a run of health issues and this slipped through. see https://reviews.freebsd.org/D14777 should land soon as its not overly complicated.

dch:
@fschulze done - thanks for the reminder

adamierymenko:
We are going to close this unfortunately. While we really appreciate the work, we think this is out of scope and should be handled by a script, systemd, or some other aspect of the system as it affects any kind of network interface (e.g. WiFi) and not just ZeroTier.

PR #453, opened 2017-03-15
Empty override means it doesn't build

jonathonf:
Having a blank override_dh_auto_build means it doesn't actually build anything. Either comment out entirely, or just remove it.


jonathonf:
Wrong destination branch... should PR into dev...

adamierymenko:
Okay, I think we'll figure out some way to make this the default.

It was done this way intentionally since in our build farm we just want to package a pre-build binary, not build in each variant.

PR #455, opened 2017-03-15
Fix empty override_dh_auto_build causing build failure

jonathonf:
A blank debian/rules directive means nothing happens. In this case, it means the build process fails. Removing the empty directive lets builds continue correctly.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -6,9 +6,6 @@ CXXFLAGS=-O3 -fstack-protector-strong
 %:
 	dh $@ --with systemd
 
-override_dh_auto_build:
-#	make -j 2
-
 override_dh_systemd_start:
 	dh_systemd_start --restart-after-upgrade
 
--- a/debian/rules.wheezy
+++ b/debian/rules.wheezy
@@ -6,6 +6,3 @@ CXXFLAGS=-O3 -fstack-protector
 %:
 	dh $@
 
-override_dh_auto_build:
-#	make -j 2
-


jonathonf:
I don't think so - running `debuild` normally cleans the build area and initiates a fresh build from the source archive. Any command run beforehand (as in the makefile change) adding files will be caught by `debuild` (e.g. 'unable to represent binary changes in source'), or those files will be cleaned during e.g. `dh_auto_clean`.

From https://github.com/zerotier/ZeroTierOne/pull/453#issuecomment-286864178 it looks like debian/rules is being used to package pre-compiled binaries, rather than compiling from source, and (AFAIK) you normally do that with just a basic directory layout and .install rules.

As a first thought, a better approach might be to enable static compilation via debian/rules, and use that build process to produce static binaries within a common package.

makersphereLabs:
@jonathonf The weird thing is, without `make one` in the Makefile, `debuild` (inside the Docker container) will attempt to install the package, instead of compiling and packaging the binaries. 😬

adamierymenko:
That was intentional. Our new build farm builds static binaries and then packages them on all the distro images so our *official* binaries can be as dependency-free as possible.

We should have our own version for this though. Might revert debian in the distro and make a debian.staticbuild or something that gets moved into place by our build script.

jonathonf:
Depending on your build hosts (e.g. if they have a common identifier) or even just an environment variable it should be possible to make it a conditional without adding extra files.

adamierymenko:
Closing since we fixed this in the next release -- we will keep alternate .static files and move them into place on our build host.

PR #462, opened 2017-03-18
use system libs for json and http-parser

fkooman:
This patch introduces the option to use the system library for json instead of the one in the "ext" directory. 

The makefile needs work, as now I hard coded the use of the system libraries (both json and http-parser), which will obviously not work for regular builds.

### DIFF
--- a/controller/EmbeddedNetworkController.hpp
+++ b/controller/EmbeddedNetworkController.hpp
@@ -39,7 +39,11 @@
 #include "../osdep/Thread.hpp"
 #include "../osdep/BlockingQueue.hpp"
 
+#ifdef ZT_USE_SYSTEM_JSON
+#include <json.hpp>
+#else
 #include "../ext/json/json.hpp"
+#endif
 
 #include "JSONDB.hpp"
 
--- a/controller/JSONDB.hpp
+++ b/controller/JSONDB.hpp
@@ -31,7 +31,13 @@
 
 #include "../node/Constants.hpp"
 #include "../node/Utils.hpp"
+
+#ifdef ZT_USE_SYSTEM_JSON
+#include <json.hpp>
+#else
 #include "../ext/json/json.hpp"
+#endif
+
 #include "../osdep/OSUtils.hpp"
 
 namespace ZeroTier {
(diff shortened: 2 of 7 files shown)


fkooman:
I also found lz4 is now embedded in the code itself. Ideally this would also be extractable.

adamierymenko:
We did this specifically to get rid of outside dependencies. Ideally we would like none.

When we attempted to ship packages that depended on lz4, http-parser, etc., we received many complaints about the package "not working" because upstream packages were broken or segfaulting because some weird Linux distribution had a slightly different version of the package.

adamierymenko:
BTW, nlohmann::json is header-only? How would this patch even work? It has no compiled library component.

adamierymenko:
The other problem we had was odd versions of libraries on some distributions causing bugs that don't exist on other distributions.

adamierymenko:
Ahh I see. It uses the system header version of the library.

We can make this kind of thing optional if it's needed for inclusion into other repositories. Most of what we do we do for our own build process and official binaries.

fkooman:
First, I think it is very good idea to have your own build for (paying) customers where you have full control over the dependencies and their versions, i.e. the `ext/` directory.

Second, for distribution packaging, what would help, is to actually put all (external) dependencies in the `ext/` directory, including lz and also specify the (API) versions of those dependencies that is required. That way packagers can figure out if their OS is up to date enough regarding the dependencies and bug their co-maintainers to update the packages. I did exactly this for Fedora/CentOS. That way there is at least a path to eventual distribution inclusion.

If some dependencies are low quality, i.e. they break API in minor version updates, have crappy release management, or are unreliable, then maybe those dependencies should not be depended on at all... A well maintained (upstream) library with timely security releases etc. should not cause issues and crashes, or maybe that particular distribution just sucks ;-)

dch:
As somebody who has spent a fair bit of time unbundling upstream projects like this, and in additional somebody who has cursed downstream projects for their inept (aka breaks my project) unbundling I feel the concerns from both sides.

It is a reasonable expectation that zerotier could be used as the primary functional network connection for a device. Having a single binary built without external dependencies is in this case is definitely a good thing. Nobody wants to go out on foot and find thousands of IoT devices to fix their networking that broke because of a dependency upgrade failure that wasn't tested by the OS distrubution.

Having the *option* to depend on platform libraries is a reasonable compromise, but considering ZT is < 1Mib size I think shipping a single no-dependencies binary is actually a big win.

PR #537, opened 2017-06-26
Update README.md

laduke:
We've had multiple questions in the community chat regarding posting new
network and getting the curl syntax correct, so adding an example.


LFlare:
I for one would love some contribution towards a [tool I've been building with Python](https://github.com/LFlare/zerotiernc), unless there's a similar tool underworks by the proper developers in C?

PR #549, opened 2017-07-08
Null ref fix

surgicalcoder:
Fixes zerotier/ZeroTierOne/#515


glimberg:
1) Merges must be against `dev`
2) This is already fixed in `dev`

PR #582, opened 2017-09-11
Java fixup/1

dharrigan:
In Java:
    
 * methods in interfaces are public by default
 * not traditionally broken at a short column length
    
Some little other tidy-ups.
    
-=david=-

### DIFF
--- a/java/src/com/zerotier/sdk/DataStoreGetListener.java
+++ b/java/src/com/zerotier/sdk/DataStoreGetListener.java
@@ -48,7 +48,6 @@ public interface DataStoreGetListener {
      * @param out_buffer buffer to put the object in
      * @return size of the object
      */
-    public long onDataStoreGet(
-            String name,
-            byte[] out_buffer);
+    long onDataStoreGet(String name, byte[] out_buffer);
+
 }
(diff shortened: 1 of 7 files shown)


glimberg:
This is nothing but cosmetic.

dharrigan:
Hi,

Thank you for your feedback. With respect, I disagree with your comment
because:

1. It removes from the interfaces an unnecessary construct that adds noise
and confusion to Java developers. Methods in Java interfaces are public by
default, so having default there is redundant and looks weird. Why keep it?

2. An additional two items of JavaDoc where added.

3. I admit, joining the lines may feel cosmetic, but what is there already
doesn't look correct to Java developers.

Whilst it is a small patch, it does no harm to merge in as it doesn't break
anything, but leaves the code a little tidier and friendlier for Java
developers. After all, adoption is good and having the code start to adhere
to Java norms should be encouraged.

I would hope, also, given my comments on the community mattermost, that
this would be the basis of other commits coming to bring the code
up-to-date for Java 8 and eventually 9.

I would be happy to resubmit the patch, removing the joined lines if you
feel very much against it.

-=david=-

PR #764, opened 2018-05-26
fix MAC address rule parsing as even/uneven switches at every colon

Elfe:
currently mac address rules do not parse correctly

drop macsrc aabbccddeeff;
drop macsrc 11:22:33:44:55:66;

both produce json with broken MACs, this patch should fix this issue

### DIFF
--- a/rule-compiler/rule-compiler.js
+++ b/rule-compiler/rule-compiler.js
@@ -226,12 +226,16 @@ function _cleanMac(m)
 {
 	m = m.toLowerCase();
 	var m2 = '';
+	let charcount = 0;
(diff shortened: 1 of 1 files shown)


glimberg:
Pull requests are only accepted against the `dev` branch.  Please close and resubmit.

adamierymenko:
Went ahead and accepted it. We will do another bug fix release soon. But in the future we should do dev... or maybe we should swap master and dev with master and release and make master dev? Not sure.

laduke:
You can edit the pull request and change the target branch (not that it wouldn't probably conflict)

PR #776, opened 2018-05-31
Improve debian service requirements

Wykks:
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=

### DIFF
--- a/debian/zerotier-one.service
+++ b/debian/zerotier-one.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=ZeroTier One
-After=network.target
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 ExecStart=/usr/sbin/zerotier-one


janjaapbos:
There are cases where zerotier is used to provide the functional network that other services depend on. Postponing zerotier is not needed I think. It should be able to handle not being online and detect when it becomes online.

Wykks:
It's just a small optimisation.

Of course, ZeroTier can start without network, and then detect when network goes up (that's why I used `Wants` instead of `Requires`). But I think, it's better if this step can be avoided most of the times.

To back up my point, in the service conf of OpenVPN server you can find
```
After=syslog.target network-online.target
Wants=network-online.target
```

> There are cases where zerotier is used to provide the functional network that other services depend on. 

Welp after reading this multiple times, I'm starting to wonder if `Before=network-pre.target` (with Wants too)
is more appropriate (https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/)


It's ok If you leave this as is, I won't bother you anymore with this :D

PR #794, opened 2018-06-19
Make sure controller.d/network/${networkId}/ exists

laduke:
FileDB tries to write to files in `controller.d/network/${networkId}/member/`
but `controller.d/network/${networkId}/` doesn't exist, so it fails.

```
controller.d/
├── network
│   ├── d7bab7f079eef5d8
│   │   └── member
│   │       └── 9935981b1e.json
│   └── d7bab7f079eef5d8.json
└── trace

```

### DIFF
--- a/controller/FileDB.cpp
+++ b/controller/FileDB.cpp
@@ -91,6 +91,8 @@ void FileDB::save(nlohmann::json *orig,nlohmann::json &record)
 				nlohmann::json old;
 				get(nwid,old);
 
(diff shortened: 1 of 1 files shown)


laduke:
Hrm, when you delete a network from the controller, the member files stick around.

After deleting the networks:
```
controller.d/
├── network
│   ├── 9935981b1e286d21
│   │   └── member
│   │       └── 9935981b1e.json

```

PR #746, opened 2018-05-13
Fix invalid JSON in network member list

michaelansel:
`/controller/network/NWID/member` is returning an invalid JSON response. This appears to have just been a minor oversight when introducing the RethinkDB integration (https://github.com/zerotier/ZeroTierOne/commit/f5014d7d7179a77311f58f8bd0dced0ea83f2885#diff-7433eba40b07c5fb84c2b5f295789d80R544)

Actual behavior:
```
curl -s -X GET --header "X-ZT1-Auth: $authtoken" http://localhost:9993/controller/network/$nwid/member
{""member1":1,""member2":1}
```

Expected behavior:
```
curl -s -X GET --header "X-ZT1-Auth: $authtoken" http://localhost:9993/controller/network/$nwid/member
{"member1":1,"member2":1}
```

I don't have a development environment set up to test this, but it _seems_ pretty straight forward (famous last words).

PR #797, opened 2018-06-25
Fix member list JSON formatting emitted from controller API

dschulz:
The controller API emits invalid JSON when querying the `controller/network/<network id>/member` endpoint. There's a single stray quotation mark that needs to be removed.


laduke:
hey thanks! is this the same as #746 ?

glimberg:
Already fixed in the dev branch

dschulz:
Indeed, it's the same as #746. Thanks and sorry for the noise :)

PR #806, opened 2018-07-04
Add Linux route metrics

neVERberleRfellerER:
Add metrics based on ZT_IF_METRIC.
There are negative implications of this - when somebody connected over zerotiers enters network that is same as one routed through managed routes, it might result in data leak as data are send to nodes in different network directly (not over zt). However, this is what allegedly happens on other platforms due to high interface metric, so this makes this Linux consistent with other platforms (who are in majority and this is what most people want anyway).
It is not marvel of great engineering, but propagating something hrough zt code is not exactly easy.
Tested on Arch and Fedora.

### DIFF
--- a/node/Constants.hpp
+++ b/node/Constants.hpp
@@ -460,4 +460,10 @@
 #define ZT_EXCEPTION_INVALID_SERIALIZED_DATA_INVALID_CRYPTOGRAPHIC_TOKEN 202
 #define ZT_EXCEPTION_INVALID_SERIALIZED_DATA_BAD_ENCODING 203
 
+// Interface and route metric for ZeroTier taps -- this ensures that
+// if we are on WiFi and also bridged via ZeroTier to the same LAN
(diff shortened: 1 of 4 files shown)


adamierymenko:
Hmm... unfortunately we've changed Linux to get rid of external calls to /sbin/ip, which changes a bunch of stuff.

What's the scenario where this was needed?

neVERberleRfellerER:
On Linux, having same network in zerotier and physical network resulted in zerotier taking precedence which broke everything as all communications were pushed through zerotier and failed since zerotier itself could not communicate due to overlap. My PR gave metrics to routes to lower priority of routes managed by zerotier. This was necessary, because Linux does not have interface metrics that serve the same purpose on other platforms. Also, this was difference between Linux and other platforms, which is confising, since overlaping networks work fine on Windows and probably OSX, but not on Linux.
I can try to rewrite this to netlink.

neVERberleRfellerER:
I'll open new PR when/if I get it done.

PR #822, opened 2018-08-02
Add functionality to erase members from networks using file backend i…

dschulz:
The FileDB.cpp was lacking the functionality to actually delete members from networks. 
This patch allows to delete the .json files corresponding to members in any given network.

Signed-off-by: Diego Schulz <dschulz@gmail.com>

### DIFF
--- a/controller/EmbeddedNetworkController.cpp
+++ b/controller/EmbeddedNetworkController.cpp
@@ -1042,6 +1042,7 @@ unsigned int EmbeddedNetworkController::handleControlPlaneHttpDELETE(
 
 					json network,member;
 					_db->get(nwid,network,address,member);
+                                        _db->eraseMember(nwid, address);
 
 					{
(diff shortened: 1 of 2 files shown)


dschulz:
This is only a partial fix. The json files get deleted right away when the DELETE method is issued to the controller api, but the member still appears when querying the member list for the same network. When the daemon is restarted after that, the member disappears, though. 

I'm not sure why this is not working as I expected. I can only guess it has something to do with the `_memberChanged()` function.

@adamierymenko can you please take a quick look and comment on this? Thanks!

adamierymenko:
This is related to #859

lpbpu:
@dschulz , the reason why member still appears when DELETE from controller api, just because misuse of DB:get function to get member's config
In FileDB::eraseMember
   get(networkId,network);
   get(memberId,member);
should changed to
   get(networkId,network,memberId,member);

PR #845, opened 2018-09-07
Using multi-stage build for Dockerfile

asbjornenge:
Hi!

I've update the Dockerfile to leverage the [mulit-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/). This way it's possible to build inside a container and still produce a small final image. That should also make it easier to automat builds by using github hooks into docker hub 👍 

Take a look and please consider automating container builds 😉 🎉


adamierymenko:
Merged to `dev` because that's where we work. Looks good.

PR #858, opened 2018-09-23
macOS uninstall script: remove symlinks from the actual location

tommyang:
Binary symlinks are in `/usr/local/bin` instead of `/usr/bin` since commit 0cf4ddd


adamierymenko:
Merged into `dev` for next release.

Issue #682, opened 2018-02-13
Need an updated zerotier image on docker hub


ghost:
current image "hangs" indefinitely when run in Kubernetes containers on aws, happens when running various commands like "netstat -a", "docker ps", etc.

OmgImAlexis:
@adamierymenko could you please setup auto building on Docker Hub.

bltavares:
I would be interested in helping with this, and also to provide support for Docker on ARM (useful for raspberry-pies)

It would be interesting to see if we could have support from https://github.com/docker-library/official-images to build a official zerotier image, and take advantage of the ARM and x86 build agents to push it there. What do you think?

rwv:
Is there any plan to update the image on [Docker Hub](https://hub.docker.com/r/zerotier/zerotier-containerized)?

bltavares:
I'm still available to help with the multi-arch official image as proposed a while ago.

adamierymenko:
Closing since it's easy enough to add it via a Dockerfile from packages and this is yet another thing for us to maintain. We're already stretched a little thin in that respect.

PR #869, opened 2018-10-21
Update ZeroTier version on Docker

bltavares:
This commit changes the Dockerfile ot use the multi-stage build and help
to get an automated build on Docker.

The idea of the multi-stage build is to use the already stable Debian
distribution channel to provide up-to-date versions of ZeroTier.

The benefit is that it would be possible to automate the image build,
either on Docker Hub, Travis or taking advantage of the [docker-library/official-images] infrastructure.

This changes follows the best-practices suggested by
[docker-library/official-images], such as using a High Availability GPG
keyserver, providing a default CMD, allowing "bash" on `docker run` and
others.

Given that both the builder `debian:stretch` and `alpine:latest` are
official messages and have [manifests], this means that this Dockerfile
is also multi-platform. This means that this same Dockerfile will
pick-up the correct Debian package according to the architecture of the
running system during build.

With this changes we could try to promote the image to be parte of
[docker-library/official-images], and take advantage of automated
publishing of multi-architecture images. Others would be able to use
`docker run zerotier` and download the latest version appropriate to
their system.

Related to #682

[docker-library/official-images]: https://github.com/docker-library/official-images
[manifests]: https://blog.docker.com/2017/09/docker-official-images-now-multi-platform/

### DIFF
--- a/ext/installfiles/linux/zerotier-containerized/Dockerfile
+++ b/ext/installfiles/linux/zerotier-containerized/Dockerfile
@@ -1,7 +1,16 @@
+FROM debian:stretch as builder
+
+## Supports x86_64, x86, arm, and arm64
+
+RUN apt-get update && apt-get install -y curl gnupg
+RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys 0x1657198823e52a61  && \
+    echo "deb http://download.zerotier.com/debian/stretch stretch main" > /etc/apt/sources.list.d/zerotier.list
+RUN apt-get update && apt-get install -y zerotier-one=1.2.12
+
 FROM alpine:latest
 MAINTAINER Adam Ierymenko <adam.ierymenko@zerotier.com>
 
-LABEL version="1.2.4"
+LABEL version="1.2.12"
 LABEL description="Containerized ZeroTier One for use on CoreOS or other Docker-only Linux hosts."
 
 # Uncomment to build in container
@@ -9,12 +18,13 @@ LABEL description="Containerized ZeroTier One for use on CoreOS or other Docker-
 
 RUN apk add --update libgcc libstdc++
 
-ADD zerotier-one /
-RUN chmod 0755 /zerotier-one
-RUN ln -sf /zerotier-one /zerotier-cli
 RUN mkdir -p /var/lib/zerotier-one
+
+COPY --from=builder /var/lib/zerotier-one/zerotier-cli /usr/sbin/zerotier-cli
+COPY --from=builder /var/lib/zerotier-one/zerotier-idtool /usr/sbin/zerotier-idtool
+COPY --from=builder /usr/sbin/zerotier-one /usr/sbin/zerotier-one
 
 ADD main.sh /
 RUN chmod 0755 /main.sh
-
-ENTRYPOINT /main.sh
+ENTRYPOINT ["/main.sh"]
+CMD ["zerotier-one"]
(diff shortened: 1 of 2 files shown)


bltavares:
I would be happy to help setup the necessary changes on [docker-library/official-images](https://github.com/docker-library/official-images#multiple-architectures) to get ZeroTier as an official image with multi-architecture support.

There is a [CHECKLIST](https://github.com/docker-library/official-images/blob/master/NEW-IMAGE-CHECKLIST.md)

Take for example [Consul](https://hub.docker.com/r/_/consul/). You can see on their Docker Hub page that they have multiple architecture supported, so any Raspberry Pi or Linux will always just have to execute `docker run consul`, without even having to specify which platform to use, nor use a prefix.

This could help promote the project, given the discoverability and how simple it would be to run on any Linux, either a x86, AMD64 or ARM. Maybe it would be simpler to manage this if a zerotier/docker project is created, given [consul's example](https://github.com/docker-library/official-images/blob/c11fd89d8e18d2005e50ba9c4dba175509048e51/library/consul#L5)

bltavares:
For example, I've made this [image](https://hub.docker.com/r/bltavares/zerotier/) multi-architecture.

Running the following both on VM or an a Raspberry Pi will provide the desired result:


```sh
docker run --device=/dev/net/tun \
    --net=host --cap-add=NET_ADMIN --cap-add=SYS_ADMIN \
    -d -v /var/lib/zerotier-one:/var/lib/zerotier-one -n zerotier-one \
    bltavares/zerotier
```

And we can see it provides multiple architectures running the following:

```sh
docker run --rm mplatform/mquery bltavares/zerotier
Image: bltavares/zerotier
 * Manifest List: Yes
 * Supported platforms:
   - linux/arm/v7
   - linux/amd64
```

bltavares:
Hi there @adamierymenko. Thanks for landing this changes.

Would it be of interest of ZeroTier to have the Docker `oficial-image` created, as my comments above?
I would be glad to contact them and get the infra setup for the multi-platform automated builds, specially if it would help spread awareness of the awesome product being built here.

I suppose that after the initial setup, it would be a matter of keeping the version on the Dockerfile updated with new versions, and update the commit hash on [docker-library/official-images](https://github.com/docker-library/official-images). I'm available to answer any questions.

hawkinchina:
@bltavares   
I use the docker images of https://hub.docker.com/r/bltavares/zerotier
it cannot work for arm64.
docker run  --device=/dev/net/tun  --net=host --cap-add=NET_ADMIN --cap-add=SYS_ADMIN -v ~/dockfile/zerotier:/var/lib/zerotier-one --name zerotier-one -d bltavares/zerotier
9bedd5f818163635d2ae2b6eba6019c55441720cb34d25ba261b3cfa82d4d9fa
docker: Error response from daemon: error gathering device information while adding custom device "/dev/net/tun": no such file or directory.

bltavares:
@hawkinchina 

Given the error message, does the Docker host have the `tun` device?

> "/dev/net/tun": no such file or directory.

Maybe it needs a package, to use `mknode` or just a restart to have the system create the missing file.

hawkinchina:
@bltavares
hi, bltavares, thanks. I used the rk3399 arm board for this docker.  I tested this docker on MacOS, it was ok, and check the /dev/net/tun again ,  it was fine.  so  I knew  maybe the linux for arm had some problem. 
the linux kernel of this board  did not have the TUN module  as default configuration while making.
Maybe i need get the source code of rk3399, set TUN module = Y ,and compile again.

thanks anyway.

PR #928, opened 2019-03-12
Label time since last send/receive in listpeers output

jstasiak:
Took me longer than it should to find out what those values meant, this
label could be useful to others.

### DIFF
--- a/one.cpp
+++ b/one.cpp
@@ -365,7 +365,7 @@ static int cli(int argc,char **argv)
 			if (json) {
 				printf("%s" ZT_EOL_S,OSUtils::jsonDump(j).c_str());
 			} else {
-				printf("200 listpeers <ztaddr> <path> <latency> <version> <role>" ZT_EOL_S);
+				printf("200 listpeers <ztaddr> <path;time in ms since last send;time in ms since last receive> <latency> <version> <role>" ZT_EOL_S);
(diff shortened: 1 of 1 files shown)


laduke:
Neat. Can you post an example of the output?

jstasiak:
Sure:

```
% ./zerotier-cli listpeers
200 listpeers <ztaddr> <path;time in ms since last send;time in ms since last receive> <latency> <version> <role>
200 listpeers 3610ad5189 - -1 1.2.12 LEAF
200 listpeers 8841408a2e 46.101.160.249/9993;3836;3754 176 1.1.5 PLANET
200 listpeers 9d219039f3 188.166.94.177/9993;3836;14395 307 1.1.5 PLANET
200 listpeers af78bf9436 35.236.10.22/44740;3644;3644 299 1.2.13 LEAF
```

I realise it's a bit on the verbose side but I chose that instead of risking being too terse. I also thought about something like this `<path;time in ms since last send;...receive>` but I'm not sure how I feel about it.

laduke:
Hmm, that would make `zerotier-cli listpeers | column -t` not line up anymore. 

(I have no control over merging here, just think this is interesting)

adamierymenko:
Might tweak a bit and merge. Thanks!

PR #932, opened 2019-04-06
Dev

mapgccv:
Dear all, for your consideration a POC of a status window for windows that lists networks and members in a user friendly way. You can enable it by checking the menu item "Status window" in the main context menu.
If it is something of interest I can refactor it to be a bit more robust and elegant - and better looking too. 
Thanks!


glimberg:
Just tried this out and it's completely broken.

1) It pops up a window on startup and we don't want that.  
2) The window doesn't populate with anythong other than "Loading networks" and shows exceptions being thrown in the console output
3) The members view is also blank and shows exceptions being thrown in the console output
4) It looks like what you're going for on this is for partial management of https://my.zerotier.com.  We're actually moving away from that for the tray application and have removed many things related to that in the Windows tray app.

PR #952, opened 2019-05-31
clang 8.0 requires -frtti on FreeBSD 13.0-CURRENT

dch:
```
node/RingBuffer.hpp:320:8: error: use of typeid requires -frtti
                        if (typeid(T) == typeid(int)) {
                            ^
node/RingBuffer.hpp:320:21: error: use of typeid requires -frtti
                        if (typeid(T) == typeid(int)) {
                                         ^
2 errors generated.
```

As usual I know nothing about what this means, *shrug* but this makes the pain go away.


joseph-henry:
Thanks for discovering this, since this issue resides in a debug function I'll just comment it out so we don't need to add `RTTI`.

PR #974, opened 2019-07-22
Don't relink zerotier-one every time.

squeakbat:
Because zerotier-one depends on "one", which is a phony target,
making it relinks zerotier-one every time, which is probably
not the desired behavior.

In any case, zerotier-one, zerotier-cli, and zerotier-idtool are
real target, so they should have real dependencies and actions.
The "one" target now depends on these three files.

Behavior is mostly unchanged otherwise.


squeakbat:
I made my change based on master.  Not sure that the correct procedure is.

Issue #738, opened 2018-05-08
1.2.8 systemd service does not exit in a timely manner


jonathonf:
## Summary

With 1.2.8, the `zerotier-one` service no longer exits in a timely manner, instead being killed after a 1m30 timeout. 1.2.6 works as expected (terminates cleanly on system halt/reboot).

## Details

Something in the [1.2.6->1.2.8 changeset](https://github.com/zerotier/ZeroTierOne/compare/1.2.6...1.2.8) is preventing the zerotier-one service from exiting in a timely manner. On system reboot or poweroff the service is killed after a 1m30 timeout.

I've tested PR #722 against 1.2.6 and it's not that bit.

I'll start bisecting but am submitting now in case there's a known/obviously suspect commit.

## Log output

```
May 07 22:00:06 box systemd[1]: Stopping ZeroTier One...
May 07 22:01:36 box systemd[1]: zerotier-one.service: State 'stop-sigterm' timed out. Killing.
May 07 22:01:36 box systemd[1]: zerotier-one.service: Killing process 1371 (zerotier-one) with signal SIGKILL.
May 07 22:01:36 box systemd[1]: zerotier-one.service: Main process exited, code=killed, status=9/KILL
May 07 22:01:36 box systemd[1]: zerotier-one.service: Failed with result 'timeout'.
May 07 22:01:36 box systemd[1]: Stopped ZeroTier One
```

## System details

OS: Manjaro Linux 64-bit
Tested kernels: 4.16.7
ZeroTier version: 1.2.8 (using packaging files from https://www.archlinux.org/packages/community/x86_64/zerotier-one/)
systemd version: 238.76

jonathonf:
That was quicker than I thought. There's a change to OneService.cpp here:

https://github.com/zerotier/ZeroTierOne/compare/1.2.6...1.2.8?diff=unified&name=1.2.8#diff-6bd361de78e54e7db32d9b4e502a88b0

After reverting this with the below patch I can shut down/reboot without systemd triggering the "kill" timer.

I can't see the __NetBSD having any effect on a Linux system (unless clang is doing something _very_ odd) so it's pointing towards the "Check global blocklists" section. I see a `Mutex::Lock` which perhaps isn't being released?


```diff
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -120,10 +120,6 @@ namespace ZeroTier { typedef WindowsEthernetTap EthernetTap; }
[...]

PR #1026, opened 2019-09-02
Fix ubuntu service to ensure Zerotier stops nicely.

AlexisTM:
As explained in #738, the Ubuntu service is not stopping properly. It was initially due to the missing `
Wants=network-online.target` but is now caused by the wrong `After` argument.

Changing the service argument from `After=network-online.target` to `After=network.target` will ensure the service to stop nicely and avoid the 90 seconds delay before Ubuntu kills Zerotier.

It was only tested on `Ubuntu 18.04`.

### DIFF
--- a/debian/zerotier-one.service
+++ b/debian/zerotier-one.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=ZeroTier One
-After=network-online.target
+After=network.target
 Wants=network-online.target
 
 [Service]


AlexisTM:
This fix doesn't work all the time. I still get to wait from time to time.

JackPShafer:
@AlexisTM Yeah, it now sometimes works and sometimes doesn't :(

PR #1031, opened 2019-09-08
Fix missing return value compile warning

### LINKED ISSUE #987: OpenBSD build hangs at 100% CPU on startup
**Describe the bug**

On first startup with clean /var/db/zerotier-one dir, zt pegs cpu at 100% and never gets going. 

- Using a debug build, no output is seen so we are getting stuck quite early.
- Only authtoken.secret and zerotier-one.pid are created.
- the port is not yet registered

```
$ uname -a
OpenBSD flatline.skunkwerks.at 6.5 GENERIC.MP#3 amd64
$ pkg_info gcc
Information for inst:gcc-4.9.4p18
 sysctl hw (abridged)
hw.machine=amd64
hw.model=Intel(R) Atom(TM) CPU C2558 @ 2.40GHz
hw.ncpu=4
hw.cpuspeed=2400
hw.physmem=8559362048
hw.usermem=8559349760
hw.smt=0
```

**To Reproduce**

- git clone zt & checkout 1.4.0 tag.
- `doas pkg_install gcc g++` & pick the 4.9 versions
- `gmake debug && doas gmake install`
- `doas zerotier-one /var/db/zerotier-one &`
- watch top in another window and cry
- check /var/db/zerotier-one for usual files

**Expected behavior**

- cpu is not pegged
- zt startup works as expected in 1.2.12

**Other info**

It would be sensible to add a couple of lines of output indicating progress during startup.
I was unable to compile zt with gcc8.3.0

ajd4096:
This fixes #987, tested on OpenBSD 6.4 amd64, OpenBSD 6.4 i386, OpenBSD 6.5 i386

### DIFF
--- a/node/Utils.hpp
+++ b/node/Utils.hpp
@@ -375,7 +375,9 @@ public:
 #if defined(__GNUC__)
 #if defined(__FreeBSD__)
 		return bswap64(n);
-#elif (!defined(__OpenBSD__))
+#elif defined(__OpenBSD__)
+		return swap64(n);
+#else
 		return __builtin_bswap64(n);
 #endif
 #else
@@ -408,7 +410,9 @@ public:
 #if defined(__GNUC__)
 #if defined(__FreeBSD__)
 		return bswap64(n);
-#elif (!defined(__OpenBSD__))
+#elif defined(__OpenBSD__)
+		return swap64(n);
+#else
 		return __builtin_bswap64(n);
 #endif
 #else


ajd4096:
The original warning was:
```
$ gmake
eg++ -O3 -fstack-protector -Wall -fPIE -fvisibility=hidden -fstack-protector -pthread  -DNDEBUG  -DZT_USE_X64_ASM_
SALSA2012 -DZT_BUILD_PLATFORM=9 -DZT_BUILD_ARCHITECTURE=2 -DZT_SOFTWARE_UPDATE_DEFAULT="\"disable\"" -std=c++11
 -c -o node/C25519.o node/C25519.cpp
In file included from node/C25519.hpp:17:0,
                 from node/C25519.cpp:15:
node/Utils.hpp: In static member function 'static uint64_t ZeroTier::Utils::hton(uint64_t)':
node/Utils.hpp:396:2: warning: no return statement in function returning non-void [-Wreturn-type]
  }
  ^
node/Utils.hpp: In static member function 'static uint64_t ZeroTier::Utils::ntoh(uint64_t)':
node/Utils.hpp:429:2: warning: no return statement in function returning non-void [-Wreturn-type]
  }

```

dch:
FWIW the 100% CPU peg bug still happens even with #1031 applied.

dch:
but +1 LGTM this at least should go in

joseph-henry:
As far as I can tell this issue has been resolved. A more generic `swapBytes` should handle the OpenBSD case so long as `__GNUC__` is defined. Going to close this without merging but if I'm wrong please re-open. As always, thanks for the effort on this.

PR #1093, opened 2019-11-28
systemd: fix zerotier hanging on shutdown

keur:
On system shutdown, zerotier is stopped after the network and gets
itself into a connection timeout loop. It hits the TimeoutStopUSec= and
is forcibly killed by SIGKILL. Instead, order zerotier after
network.target so it can shutdown gracefully while the network is still
up.

From systemd.special(7):

at shutdown, a unit that is ordered after network.target will be stopped
before the network — to whatever level it might be set up then — is shut
down. It is hence useful when writing service files that require network
access on shutdown, which should order themselves after this target, but
not pull it in

### DIFF
--- a/debian/zerotier-one.service
+++ b/debian/zerotier-one.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=ZeroTier One
-After=network-online.target
+After=network-online.target network.target
 Wants=network-online.target
 
 [Service]


HaleTom:
I'd suggest the following additions:

```
[Unit]
# Wait for network online (service not activated by default) before starting
After=NetworkManager-wait-online.service

# Shutdown zerotier before tearing down the network
# https://github.com/zerotier/ZeroTierOne/pull/1093/commits/c9f07e855e2abac9c1a29cd412d888500a6a0bbb
After=network.target

[Service]
# Override default 90 second timeout
TimeoutStopSec=5
```

keur:
@HaleTom Thanks. But it looks like they are going to let this PR sit forever.

marcosx86:
Confirming that happens on Debian 9 and 10

laduke:
In the mean time, between releases
 
`sudo systemctl edit zerotier-one --full`


https://www.digitalocean.com/community/tutorials/how-to-use-systemctl-to-manage-systemd-services-and-units#editing-unit-files

thurstylark:
I am also having this issue on Arch. 

I would not use `After=NetworkManager-wait-online.service` as suggested because NM isn't guaranteed to be in use on every system

I *would* add `TimeoutStopSec=5` as there shouldn't be any reason for ZT-O to hold up boot for more than that long. If there is, this can be overridden by the system admin.

Will test these options out and report back.

HaleTom:
@thurstylark I tested on my system:

```
systemctl disable --now NetworkManager-wait-online.service
```

Things still start fine with the line:

    After=NetworkManager-wait-online.service

It's an ordering, not a dependency ([man page](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)):

> Note that those settings are independent of and orthogonal to the requirement dependencies as configured by Requires=, Wants=, Requisite=, or BindsTo=. It is a common pattern to include a unit name in both the After= and Wants= options, in which case the unit listed will be started before the unit that is configured with these options.

So the line

    After=NetworkManager-wait-online.service

**should** be there, *in case* `NetworkManager-wait-online.service` is enabled.

HaleTom:
@keur Given the above, would you please update the `After` line to be:

```
After=network-online.target network.target NetworkManager-wait-online.service
```

traverseda:
Is this something we can get merged in? It would be nice to not have to edit this file for every device.

Issue #990, opened 2019-08-03
Packet characteristic "ipauth" reports false for addresses within 6plane


fakuivan:
**Describe the bug**
The packet characteristic "ipauth" reports false for source addresses within the address space assigned by 6plane to that node. This means that if you choose to drop unauthorized packets, you will also be _disabling_ 6plane for any address different from the one assigned to the node itself.

**To Reproduce**
1. Create a Network with the default rule set and uncomment the ``drop not chr ipauth;`` line
2. Enable 6plane addressing
3. Add two nodes, alice and bob
4. Follow [this guide](https://zerotier.atlassian.net/wiki/spaces/SD/pages/7274520/Using+NDP+Emulated+6PLANE+Addressing+With+Docker) to configure Docker to use 6plane for all containers on bob
4. Ping alice with the command ``docker run --rm -it alpine ping "<alice's 6plane address>"``
5. Monitor traffic using ``tcpdump 'icmp6 && (ip6[40] == 128 || ip6[40] == 129)' -i "<zt network interface>"`` on both alice and bob
6. Notice that packets do leave the container and bob but do not reach alice. 

This is even more evident if you try to ping the container from alice, the packet will be received by bob, relayed to the container and vice versa, but once it leaves bob to reach alice, it will get silently dropped by zt. 

**Expected behavior**
A single node should be able to relay packets from the 6plane addressing space assigned to it, packets sent from there should be identified as authorized. 

**Screenshots**
![2019-08-02_19-35-42](https://user-images.githubusercontent.com/15100012/62405459-be4b8e80-b573-11e9-963c-92f0f5924f39.gif)

**Desktop (please complete the following information):**
 - OS: Linux, Windows
 - OS/Distribution Version: Ubuntu Server 18 LTS, Windows 10
 - ZeroTier Version: 1.2.4, 1.4
 - Hardware: Virtual Machines hosted with Hyper-V, same behaviour observed on bare metal servers hosting docker containers
 
**Smartphone (please complete the following information):**
 - Device: PC
 - OS: Linux, Windows
 - Version: 1.4.0, 1.2.12

**Additional context**
- General Network Environment: Hyper-V internal network, home LAN

adamierymenko:
You know what... I'm not sure we have ever tried this! Technically this mechanism is redundant for 6PLANE and similar addresses as they embed the ZT address into the IP itself, emulate NDP, and therefore cannot be spoofed. But it should still work in the rules engine the way you would expect.

[...]

PR #1123, opened 2020-01-09
Fix permissive filter and check 6plane for outbound packets

fakuivan:
As far as I understand it, the changes made in 52a166a to attempt to solve #990 made it so that _any_ NDP emulated address would count as valid source for an inbound packet, regardless of the sender node. 
Also, the output of ``_isV6NDPEmulated`` is only taken into account when marking inbound packets as ``ZT_RULE_PACKET_CHARACTERISTICS_SENDER_IP_AUTHENTICATED``, outbound packets from the 6plane /80 subnet (other than ::1) would not get this this flag and thus dropped by the sender before they can get validated by the reciever as authentitcated ([node/Network.cpp#387](https://github.com/zerotier/ZeroTierOne/compare/master...fakuivan:master#diff-ac28ce1f807c26308d3afa300bc5bee5L387)), I'm not sure the changes I made on that file are valid, so please provide feedback on this if possible.

### DIFF
--- a/node/Membership.hpp
+++ b/node/Membership.hpp
@@ -181,38 +181,21 @@ private:
 	inline bool _isV6NDPEmulated(const NetworkConfig &nconf,const MAC &m) const { return false; }
 	inline bool _isV6NDPEmulated(const NetworkConfig &nconf,const InetAddress &ip) const
 	{
-		if ((ip.isV6())&&(nconf.ndpEmulation())) {
-			const InetAddress sixpl(InetAddress::makeIpv66plane(nconf.networkId,nconf.issuedTo.toInt()));
-			for(unsigned int i=0;i<nconf.staticIpCount;++i) {
-				if (nconf.staticIps[i].ipsEqual(sixpl)) {
-					bool prefixMatches = true;
-					for(unsigned int j=0;j<5;++j) { // check for match on /40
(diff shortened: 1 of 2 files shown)


someara:
I'm doing a PR sweep... @adamierymenko or @joseph-henry what should we do about this 3 year old PR?

node/Membership.hpp
```diff
@@ -181,38 +181,21 @@ class Membership
...
+		const InetAddress rfc4193(InetAddress::makeIpv6rfc4193(nconf.networkId, nconf.issuedTo.toInt()));
+		for (unsigned int i = 0; i < nconf.staticIpCount; ++i) {
+			bool prefixMatches = false;
+			// For 6plane addresses we only need to check the first 80 bits, since 6plane is mainly
+			// designed to delegate that address space to docker containers running on that node
+			if (nconf.staticIps[i].ipsEqual(sixplane)) {
+				if (sixplane.ipsEqual(ip)) return true;
+				for (unsigned int j = 0; j < 10; ++j) {	// check for match on /80
+					if (prefixMatches = (((const struct sockaddr_in6 *)&ip)->sin6_addr.s6_addr)[i] !=
+						(((const struct sockaddr_in6 *)&sixplane)->sin6_addr.s6_addr)[i]) break;
```
fakuivan, line 195: I'm not completely sure this does what's supposed to, `j` is never used

PR #1159, opened 2020-03-14
Add support for MidnightBSD. Tested on 1.2 stable.

laffer1:


adamierymenko:
1.2 is really, really ancient, and V2 will use cmake.

Issue #1213, opened 2020-06-04
Error In Compile Zerotier 


OpenVu:
Hello

I get this Error in compile zerotier For My setop Box with Gcc 10 

===Log====
 In file included from service/../node/Membership.hpp:24,
|                  from service/../node/Network.hpp:37,
|                  from service/../node/Node.hpp:32,
|                  from service/OneService.cpp:33:
| service/../node/Capability.hpp: In constructor 'ZeroTier::Capability::Capability()':
| service/../node/Capability.hpp:70:37: warning: 'void* memset(void*, int, size_t)' clearing an object of type 'struct ZeroTier::Capability::<unnamed>' with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
|    70 |   memset(_custody,0,sizeof(_custody));
|       |                                     ^
| In file included from service/../node/Membership.hpp:24,
|                  from service/../node/Network.hpp:37,
|                  from service/../node/Node.hpp:32,
|                  from service/OneService.cpp:33:
| service/../node/Capability.hpp:476:9: note: 'struct ZeroTier::Capability::<unnamed>' declared here
|   476 |  struct {
|       |         ^
| osdep/OSUtils.cpp: In static member function 'static std::vector<std::__cxx11::basic_string<char> > ZeroTier::OSUtils::listDirectory(const char*, bool)':
| osdep/OSUtils.cpp:111:28: warning: 'int readdir_r(DIR*, dirent*, dirent**)' is deprecated [-Wdeprecated-declarations]
|   111 |   if (readdir_r(d,&de,&dptr))
|       |                            ^
| In file included from osdep/OSUtils.cpp:32:
| /home/enigma/vision/openvision-development-platform/build/tmp/work/mips32el-oe-linux/zerotier/1.4.6+gitAUTOINC+b6b11dbf82-r0/recipe-sysroot/usr/include/dirent.h:183:12: note: declared here
|   183 | extern int readdir_r (DIR *__restrict __dirp,
|       |            ^~~~~~~~~
| osdep/OSUtils.cpp: In static member function 'static long int ZeroTier::OSUtils::cleanDirectory(const char*, int64_t)':
| osdep/OSUtils.cpp:161:28: warning: 'int readdir_r(DIR*, dirent*, dirent**)' is deprecated [-Wdeprecated-declarations]
|   161 |   if (readdir_r(d,&de,&dptr))
|       |                            ^
[...]

PR #1232, opened 2020-07-14
Fix compilation with GCC 10 by providing explicit cast of _bindingCount

diabonas:
Compilation with GCC 10 fails with the following error:
```
service/../osdep/Binder.hpp: In member function ‘void ZeroTier::Binder::refresh(ZeroTier::Phy<PHY_HANDLER_TYPE>&, unsigned int*, unsigned int, std::vector<ZeroTier::InetAddress>, INTERFACE_CHECKER&)’:
service/../osdep/Binder.hpp:376:30: internal compiler error: unexpected expression ‘(std::__atomic_base<unsigned int>::__int_type)((ZeroTier::Binder*)this)->ZeroTier::Binder::_bindingCount’ of kind implicit_conv_expr
  376 |       _bindings[_bindingCount].udpSock = udps;
      |                              ^
```
Help the compiler by providing an explicit cast to the appropriate type like it is already done in [l. 338 of the same file](https://github.com/zerotier/ZeroTierOne/blob/91b16310ea47a6de96edb488a61494f8ed8c139c/osdep/Binder.hpp#L338).

Fixes: #1213

### DIFF
--- a/osdep/Binder.hpp
+++ b/osdep/Binder.hpp
@@ -373,9 +373,9 @@ public:
 					}
 #endif // __LINUX__
 					if (_bindingCount < ZT_BINDER_MAX_BINDINGS) {
-						_bindings[_bindingCount].udpSock = udps;
-						_bindings[_bindingCount].tcpListenSock = tcps;
-						_bindings[_bindingCount].address = ii->first;
+						_bindings[(unsigned int)_bindingCount].udpSock = udps;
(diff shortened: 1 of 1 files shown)


neheb:
Can confirm this fixes it.

diabonas:
This issue does not occur for me any more using GCC 10.2.0 (it did with GCC 10.1.0), so I am closing this PR.

PR #1288, opened 2020-11-07
FreeBSD ARMv6, ARMv7, and Aarch64 support

darkain:
Fixing builds on FreeBSD

* Added linker flags to fix arm32-neon-salsa2012-asm
* Added ZT_AES_NO_ACCEL to ARMv7 (this could possibly be improved, but I think it'll require changes in the FreeBSD source tree)
* Added -march=armv8-a+crypto to ARM64 (without this, clang refused to recognize crypto intrinsic)
* Added a shim for getauxval since this isn't available on FreeBSD
* The HWCAP defines and detection could probably be cleaned up / improved, but this at least works

### DIFF
--- a/make-bsd.mk
+++ b/make-bsd.mk
@@ -80,16 +80,16 @@ ifeq ($(CC_MACH),armv6kz)
 endif
 ifeq ($(CC_MACH),armv7)
 	ZT_ARCHITECTURE=3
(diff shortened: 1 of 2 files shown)


dch:
@darkain awesome, please @ me for any future patches!

PR #1304, opened 2020-11-26
BSD: add debug fprintf for all external commands

dch:
Clearly there will be a better way to wrap/handle this from the
C++ side, if there's a preferred approach please give me an example.

I'll be pushing a few more patches to this as I find the last few `execl`
calls.

### DIFF
--- a/osdep/BSDEthernetTap.cpp
+++ b/osdep/BSDEthernetTap.cpp
@@ -106,6 +106,9 @@ BSDEthernetTap::BSDEthernetTap(
 		if (std::find(devFiles.begin(),devFiles.end(),std::string(tmpdevname)) == devFiles.end()) {
 			long cpid = (long)vfork();
 			if (cpid == 0) {
+#ifdef ZT_TRACE
+				fprintf(stderr, "DEBUG: ifconfig %s create" ZT_EOL_S, tmpdevname);
+#endif
 				::execl("/sbin/ifconfig","/sbin/ifconfig",tmpdevname,"create",(const char *)0);
 				::_exit(-1);
 			} else if (cpid > 0) {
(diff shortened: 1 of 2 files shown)


dch:
@darkain I'm considering adding this as a patch to the FreeBSD port. Do you think there's a better way than my changes above to do this? Would it be a useful debugging addition? It only appears if you run ZT in the foreground, obviously.

darkain:
I think there should be either a compile-time or run-time flip to enable/disable the extra verbosity. Compile time with macros, or run time with either a config item or command line item to add them in. Because otherwise, all of that verbosity may end up in log files instead, which could be a distraction from general logging.

joseph-henry:
Thanks for this, guys.

I'd suggest gating it as such:

```
#ifdef ZT_TRACE
...
#endif
```

Once we've done that I will go ahead and merge the request if it's still needed.

dch:
@joseph-henry done, and yes these are needed - thanks!

dch:
BTW it would be *much* more convenient if we could have these debug flags compiled in by default, but only displayed by toggling an env var or something. Rebuilding is not really practical for end users.

dch:
Regarding making it toggle-able without recompiling there are a few options:

- add a commandline flag or environment var, set a global var, and check that before displaying
- add a setting in local.conf, which would allow enabling debugging without daemon restart. as state changes (tap interfaces, route propagations) are relatively rare, this would not slow down the "hot path"

joseph-henry:
Looks great. Thanks for the modification. And yes I do agree that this should be compiled in and controlled via a flag. We're in the planning stages for a much more detailed logging system so it won't be like this forever.

Issue #1317, opened 2020-12-08
Fails to compile for ARMv8


RookieZoe:
Got an error when build openwrt with zerotier module. (build by github actions: [action](https://github.com/RookieZoe/openwrt-builder/runs/1508388547?check_suite_focus=true))

```
cc1plus: error: invalid feature modifier in '-march=armv8-a+aes+crypto'
```

I guess the feature modifier is '-march=armv8-a+crypto' instead of '-march=armv8-a+aes+crypto'. 

I found some docs on [https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gcc/AArch64-Options.html](https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gcc/AArch64-Options.html)

> Feature crypto implies aes, sha2, and simd, which implies fp. Conversely, nofp implies nosimd, which implies nocrypto, noaes and nosha2.

I'm not sure though.

## Additional information 

**Desktop (please complete the following information):**
 - OS: Linux
 - OS/Ubuntu 18.04
 - ZeroTier Version 1.6.2

## Error logs:

```
make[3]: Entering directory '/home/runner/work/openwrt-builder/openwrt-builder/openwrt/feeds/packages/net/zerotier'
mkdir -p /home/runner/work/openwrt-builder/openwrt-builder/openwrt/dl
SHELL= flock /home/runner/work/openwrt-builder/openwrt-builder/openwrt/tmp/.zerotier-1.6.2.tar.gz.flock -c '  	/home/runner/work/openwrt-builder/openwrt-builder/openwrt/scripts/download.pl "/home/runner/work/openwrt-builder/openwrt-builder/openwrt/dl" "zerotier-1.6.2.tar.gz" "c8087b26c1191d36fda004b42cdfed31042cafd8586e49015586eef786f2c9a5" "" "https://codeload.github.com/zerotier/ZeroTierOne/tar.gz/1.6.2?"    '
rm -f /home/runner/work/openwrt-builder/openwrt-builder/openwrt/build_dir/target-aarch64_generic_musl/ZeroTierOne-1.6.2/.built
touch /home/runner/work/openwrt-builder/openwrt-builder/openwrt/build_dir/target-aarch64_generic_musl/ZeroTierOne-1.6.2/.built_check
[...]

PR #1331, opened 2020-12-21
fix issue [#1317] [Fails to compile for ARMv8]

RookieZoe:
as titled.


adamierymenko:
Merged into `dev` manually, thanks!

PR #1347, opened 2021-02-04
Patch 1 - Language suggestion

showipintbri:
Removed some language from the comments in the *.inf files. 

Example: "fuqdat"


joseph-henry:
Thanks for polishing things up!

PR #1354, opened 2021-02-19
fix sprintf overflow

contrun:
Here sprintf will always overflow as the length of macStr is 17.


joseph-henry (approved):
Thank you for spotting this!

PR #1355, opened 2021-02-24
Initial docker release image tooling

erikh:
Here's some docker release tooling!

I'll document it more once it boils down to something we want to use, but run it like so:

```
docker build -f Dockerfile.release --build-arg VERSION=1.6.4 -t zerotier .
docker run -it --privileged -v zt:/var/lib/zerotier --device /dev/net/tun zerotier <networkA> <networkB>
```

Identities probably need to be resolved still in a cleaner way, but you should
be able to mount them into that volume and it should work.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,0 +1,22 @@
+# vim: ft=dockerfile
+
+FROM debian:buster as stage
+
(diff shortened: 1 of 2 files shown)


someara (approved):
Thanks Erik!

PR #1364, opened 2021-03-09
Show the port number which is not able to connect

Pablohn26:
For users that do not know/remember the default port and for better error explanation I add to the error message the port number to know which is the port that bind process is not able to get.


joseph-henry:
Good call, thanks.

PR #1369, opened 2021-03-16
Corrects crypto code debug build objects not being C++11

ALTinners:


joseph-henry:
Thanks. Which compiler was complaining about this?

PR #1377, opened 2021-03-26
Fix compilation error on MSVC and refactor code

GermanAizek:
By default, windows.h defines `min` and `max` as macros. When those are expanded, code that tries to use `std::min` (for example) will end up looking something like this:

```cpp
int k = std::(x) < (y) ? (x) : (y);
```

The error message is telling you that `std::(x)` isn't allowed.

### DIFF
--- a/include/ZeroTierOne.h
+++ b/include/ZeroTierOne.h
@@ -23,6 +23,7 @@
 
 // For the struct sockaddr_storage structure
 #if defined(_WIN32) || defined(_WIN64)
+#define NOMINMAX
 #include <WinSock2.h>
 #include <WS2tcpip.h>
 #include <Windows.h>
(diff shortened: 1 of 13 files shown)


GermanAizek:
Fixed possible memleaks, code in some places optimized.

GermanAizek:
Note but this commit is very important
https://github.com/zerotier/ZeroTierOne/pull/1377/commits/db7aef1a61022faf5329ef3ff5f1dea5a04d4c6b

GermanAizek:
@nelsonjchen @laduke @erikh @dnadlinger 
I am waiting to review pull request and your opinion.

GermanAizek:
I forgot that I was already re-creating another PR that was correctly composed, sorry.

GermanAizek:
Reference:
https://github.com/zerotier/ZeroTierOne/pull/1378/

PR #1382, opened 2021-04-08
Fix Synology compilation issue.  Looks like a bad merge lost this change

Ranbato:
Making this change allows me to build 1.6.4 version of Synology and so far seems to work on my NAS

### DIFF
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -1983,7 +1983,7 @@ public:
 				}
 			}
 #ifdef __SYNOLOGY__
-			if (!n.tap->addIpSyn(newManagedIps))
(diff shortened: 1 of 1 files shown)


Ranbato:
Been testing for almost a week now and haven't discovered any issues.  Unfortunately, I've been forced to upgrade to DSM 7.0 beta and it won't run under that so I can't continue to test.

joseph-henry:
Since the supported method of running ZT on a Synology has changed with the DSM 7 update (See: [docs.zerotier.com/devices/synology](https://docs.zerotier.com/devices/synology)) I'm going to close this. But we do appreciate the attempt to fix it.

Issue #1387, opened 2021-04-19
Using Zerotier DNS causes Network Issues in Google Chrome on macOS


robertschreib:
## Required

- What you expect to be happening.
Stable Browsing Experience in Google Chrome
- What is actually happening?
Every Time the DNS servers get synced, I get a `ERR_NETWORK_CHANGED` in Chrome.
- Any steps to reproduce the error.
Set up a new ZeroTier network with a DNS resolver configured. Every time ZeroTier synchronizes it's DNS Servers Chrome will produce an `ERR_NETWORK_CHANGED`, or any running Downloads will fail with the Error  `Network disconnected`.

## Additional information 

**Desktop (please complete the following information):**
 - macOS 10.15.7
 - ZeroTier Version 1.6.4

joseph-henry:
This appears to have been fixed. Thanks for your PR. Closing.

PR #1388, opened 2021-04-21
Fix macOS DNS issue

robertschreib:
This should Fix https://github.com/zerotier/ZeroTierOne/issues/1387


rburgst:
@glimberg when can we get a release with this fix?

PR #1408, opened 2021-06-01
Small dockerfile release fixes

erikh:
- Activate zerotier-one.port file (we'll do more with this later)
- Add a better process trap

Signed-off-by: Erik Hollensbe <linux@hollensbe.org>


laduke:
works on my machine

PR #1410, opened 2021-06-07
Add support for auto-fetched orbit like that network.d is done

SalimTerryLi:
# What does this PR change

Since networks can be automatically joined by touching an empty file under `networks.d/`, this PR makes the moons work in the same way:

1. add corresponding file under `moons.d/` with contents of target root id
2. launch `zerotier-one` service and will automatically orbit

Multiple root ID is supported by adding multiple lines.

# Who will benefit from this

- someone who run ZT on OpenWRT with no proper way to automatically orbit (UCI)

# How it is implemented

1. read `xxxxxxxxxxxxxxxx.moon` at first
2. split it into lines
3. check if all lines are valid u64 hex val or just empty line:

#### TRUE

4. delete that ASCII file and orbit those lines of moon
5. done

#### FALSE

4. give up and treat the file as serialized object
5. ZT will orbit it

# Additional info

#### If `xxx.moon` is an empty file?

This implementation treats it as orbit no moon, as no root ID provided alongside world ID

#### If `xxx.moon` is not a valid format to be recognized by this PR?

Then it works the same way as before.

### DIFF
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -797,38 +797,38 @@ public:
 				for(std::vector<std::string>::iterator f(moonsDotD.begin());f!=moonsDotD.end();++f) {
 					std::size_t dot = f->find_last_of('.');
 					if ((dot == 16)&&(f->substr(16) == ".moon")){
-					    // First try to decode as ASCII configuration
-					    bool successParsedAsASCII = true;
-                        std::string root_list_raw;
-                        std::vector<std::string> root_list;
-                        std::vector<std::string> root_list_filterd;
-                        if(OSUtils::readFile((_homePath + ZT_PATH_SEPARATOR_S "moons.d" ZT_PATH_SEPARATOR_S + *f).c_str(), root_list_raw)){
-                            root_list = OSUtils::split(root_list_raw.c_str(),ZT_EOL_S, nullptr, nullptr);
-                            // Verify each line
-                            for(auto & root_it : root_list){
-                                if (root_it.length() == 0)
-                                    continue;   // Skip empty line
-                                if (root_it.length() <= 16){
-                                    if (std::all_of(root_it.begin(), root_it.end(), [](char c){return std::isxdigit(c);})){
-                                        root_list_filterd.push_back(root_it);
-                                        continue;   // Accept valid line
-                                    }
-                                }
-                                // Once invalid line appears, break
-                                successParsedAsASCII = false;
-                                break;
-                            }
-                        }
-                        if (!successParsedAsASCII){
-                            // Feed to deserializer
-                            _node->orbit((void *)0,Utils::hexStrToU64(f->substr(0,dot).c_str()),0);
-					    }else{
-                            // Remove ASCII configuration
-                            OSUtils::rm((_homePath + ZT_PATH_SEPARATOR_S "moons.d" ZT_PATH_SEPARATOR_S + *f).c_str());
-                            for(auto & root_it : root_list_filterd){
-                                _node->orbit((void *)0,Utils::hexStrToU64(f->substr(0,dot).c_str()),Utils::hexStrToU64(root_it.c_str()));
-                            }
-                        }
+						// First try to decode as ASCII configuration
+						bool successParsedAsASCII = true;
+						std::string root_list_raw;
+						std::vector<std::string> root_list;
+						std::vector<std::string> root_list_filterd;
+						if(OSUtils::readFile((_homePath + ZT_PATH_SEPARATOR_S "moons.d" ZT_PATH_SEPARATOR_S + *f).c_str(), root_list_raw)){
+							root_list = OSUtils::split(root_list_raw.c_str(),ZT_EOL_S, nullptr, nullptr);
+							// Verify each line
+							for(auto & root_it : root_list){
+								if (root_it.length() == 0)
+									continue;   // Skip empty line
+								if (root_it.length() <= 16){
+									if (std::all_of(root_it.begin(), root_it.end(), [](char c){return std::isxdigit(c);})){
+										root_list_filterd.push_back(root_it);
+										continue;   // Accept valid line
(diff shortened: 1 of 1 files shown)


SalimTerryLi:
Find no code style related things so didn't format the file yet.

glimberg:
I'm confused as to the purpose of your pull request here.  .moon files are already automatically processed on startup of Zerotier.

https://github.com/zerotier/ZeroTierOne/blob/6faca86bb424d0b9643b6efa50571f73310d8276/service/OneService.cpp#L794

glimberg:
Additionally, it should not be removing the configuration file upon success.  things under ZeroTier's home directory are ZeroTier's state.  Removing the moon file on success removes the ability to re-orbit the moon on next startup.

glimberg:
Yeah.  I'm going to go ahead and close this request.  The behavior already exists given a properly structured & formatted `.moon` file is already in place on the filesystem.

SalimTerryLi:
> Yeah. I'm going to go ahead and close this request. The behavior already exists given a properly structured & formatted `.moon` file is already in place on the filesystem.

No, no. You definitely missed what I want it behaves.

ZT will only orbit the moon if it is a well-structured `.moon` file, but will join the network even if it is an empty file. That's what I mean 'different'. 

On systems like OpenWRT, all those ZT 'states' is not permanent at all. So, `.moon` files got deleted each time the device reboots. We can easily touch an empty file and keep the secret to let ZT on OpenWRT keep its ID and automatically join the network on every restart, but cannot orbit anything reliable. It is hard to fetch a well-structured `.moon` by hand so I came up this this idea, that tell ZT the worldID and rootID and let ZT itself fetch the necessary data.

glimberg:
That's the concern of the OpenWRT implementation, not this package.  Much like they place the identity files in the proper location with he proper content on OpenWRT startup, they can do the same with the moon files.

SalimTerryLi:
> That's the concern of the OpenWRT implementation, not this package. Much like they place the identity files in the proper location with the proper content on OpenWRT startup, they can do the same with the moon files. Some implementation work on the OpenWRT side may need to be done to get it working with moons, however.

That because secret ID is a human-readable string, but `.moon` is not.

glimberg:
OpenWRT has a base64 package (https://openwrt.org/packages/pkgdata/coreutils-base64).  The content could be base64 encoded into the configuration file & then base64 decoded into the file content on startup.

SalimTerryLi:
> OpenWRT has a base64 package (https://openwrt.org/packages/pkgdata/coreutils-base64). The content could be base64 encoded into the configuration file & then base64 decoded into the file content on startup.

It will be a lot to explain why it is not practical and you may also not interested in so I want to discuss from another side.

The implementation of this solution only works differently if the moon file contains valid root ID. If there is a single line not apply to the following rule:

- Is bigger than any possible root ID
- contains anything not within `[0-9a-fA-F\n]`

Then the `.moon` file is handed over to already existed logic.

It is a very strict rule with almost no exception.

Kind of 'hack', but not breaking any current behavior.

#### When the ASCII version of `.moon` will be delected

The file is actually an ASCII file contains valid root ID. Then ZT will orbit that moon exactly.

#### When the ASCII version of `.moon` will lost but ZT not successfully orbit

ZT crashed or system went down right after the file is deleted but ZT haven't write the true `.moon` out. 
But that will affect nothing as:

- On OpenWRT the ASCII version of moon will be created again when ZT is going to be started
- On normal system with `zerotier-cli orbit` command: `.moon` never get deleted

glimberg:
We're probably not going to accept any patch like this unless it's fixing a bug in the current implementation.  Your patch does not, and it changes the behavior of how moons work.   I've given you suggestions on ways to implement this on the OpenWRT side by placing the moon file in the proper location before starting ZeroTier.  That is and will be supported for the remainder of 1.x.  

As stated before, you can base64 encode the moon file content such that it can be stored in OpenWRT's text configuration files, then use the startup scripts to decode & place it at the appropriate place in the filesystem before ZeroTier starts.  The same way that is done with the identity on OpenWRT.  This is all the concern of the OpenWRT port, however, and doesn't need to be handled here in the main ZeroTier project.  ZeroTier is designed to keep its state between restarts.  I understand OpenWRT doesn't do that, but they have developed ways to handle that.  I suggest working within those already developed systems to handle the state on the OpenWRT side.

SalimTerryLi:
> We're probably not going to accept any patch like this unless it's fixing a bug in the current implementation. Your patch does not, and it changes the behavior of how moons work. I've given you suggestions on ways to implement this on the OpenWRT side by placing the moon file in the proper location before starting ZeroTier. That is and will be supported for the remainder of 1.x.
> 
> As stated before, you can base64 encode the moon file content such that it can be stored in OpenWRT's text configuration files, then use the startup scripts to decode & place it at the appropriate place in the filesystem before ZeroTier starts. The same way that is done with the identity on OpenWRT. This is all the concern of the OpenWRT port, however, and doesn't need to be handled here in the main ZeroTier project. ZeroTier is designed to keep its state between restarts. I understand OpenWRT doesn't do that, but they have developed ways to handle that. I suggest working within those already developed systems to handle the state on the OpenWRT side.

We have ways to handle that state by encoding into base64. But we have no way to get the initial content of the moon with OpenWRT itself. Only if:

1. We write a lot and wrap the `zerotier-cli orbit` command with another helper function
2. notify the user when it is possible to orbit
3. watch and see if any new moon appears and then store them with base64.

Quite annoying and, still cannot:

- configure which moon to orbit before ZT already started for the first time on each device.

That is what we need. I'll try to contact with openwrt side to see if it is possible to accept this solution as a every-time patch.

glimberg:
Correct.  It's something you'll have to plan for.  Much like you have to plan for the network ID you wish to join.  Again, your request doesn't belong in this project, but should be handled on OpenWRT's side.

PR #1421, opened 2021-06-25
cleaning up docker entryoint logic and disusing nohup

someara:
Test logic -
-a and -o have been marked obsolescent - https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

nohup:
ZeroTier -d forks.. this will allow debug and trace info to show up in docker logs
https://github.com/zerotier/ZeroTierOne/blob/6faca86bb424d0b9643b6efa50571f73310d8276/one.cpp#L2259

### DIFF
--- a/entrypoint.sh.release
+++ b/entrypoint.sh.release
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 grepzt() {
-  [ ! -n "$(cat /var/lib/zerotier-one/zerotier-one.pid)" -a -d "/proc/$(cat /var/lib/zerotier-one/zerotier-one.pid)" ]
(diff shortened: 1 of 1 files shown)


entrypoint.sh.release
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 grepzt() {
-  [ ! -n "$(cat /var/lib/zerotier-one/zerotier-one.pid)" -a -d "/proc/$(cat /var/lib/zerotier-one/zerotier-one.pid)" ]
+  [ -n "$(cat /var/lib/zerotier-one/zerotier-one.pid)" ] && [ -d "/proc/$(cat /var/lib/zerotier-one/zerotier-one.pid)" ]
```
erikh, line 4: you have inverted this logic, I think? is it really detecting if the service is up now?

PR #1424, opened 2021-06-29
add null check

navhaxs:
## Summary

On my Windows system, the tray app more than often shows "Node ID: unknown". This PR completely fixes this issue for me.

## Explanation

I did some digging and found that a missing null check is causing the monitor update thread to die silently.
At least on my system, the `apiStatusCallback` gets fired **before** `SubscribeStatusUpdates`. With the null check added, the thread doesn't die and this issue is fixed.

## Issue screenshot (before fix)

image

After applying the PR, the Node ID and the networks list all works as expected :)

### DIFF
--- a/windows/WinUI/NetworkMonitor.cs
+++ b/windows/WinUI/NetworkMonitor.cs
@@ -102,10 +102,10 @@ namespace WinUI
 
         private void apiNetworkCallback(List<ZeroTierNetwork> networks)
         {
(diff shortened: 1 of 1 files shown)


joseph-henry:
Thanks for this improvement. Unfortunately this change will be obsoleted by an upcoming update to the UI in 1.8. I'm going to close this without merging but we still value your effort here.

PR #1443, opened 2021-09-03
Update PostgreSQL.cpp

sbilly:
config["dns"] is not must have configuration.


glimberg:
already handled in another branch.

glimberg:
Also note: this code is not included in the downloadable binaries of ZeroTier.  While we do ship the source, it is unsupported for end users.

Issue #750, opened 2018-05-16
[Feature Request] my.zerotier.com hability to set route metric to static routes


leleobhz:
Hello folks!

Today in my.zerotier.com, we are able to setup static routes in the managed routes screen. But this place does not allow setup route metric. This way, as example, if I connect to ZT network that have a route to the network i'm joined, the route will not work because ZT usually haves a lower metric than local interface.

Can you please add this feature?

Thanks!

glimberg:
I'm not quite sure what you're asking here.

Do you have a physical network and a ZeroTier network where the IP address ranges overlap?  Because if that's the case, then that's a good way to have problems.   If this is the case, your best bet is to make sure that your IP address ranges don't overlap.  

Additionally, not all platforms allow setting route metrics (iOS and Android for instance).

novirium:
I _think_ this is the same as the issue I'm having it my situation: I've got a ZT network being used to connect several nodes including roaming laptops together, and have a gateway node set up in the office with a managed route to direct traffic to the office subnet from the ZT one. This works great until a laptop (running Linux here) connects to the office network directly, and the managed route added by ZeroTier has higher priority than the direct LAN one. In this situation, internet access is via the LAN subnet - which the zerotier route is overriding, resulting in a complete loss of internet connectivity (except to other hosts on the ZT subnet, which still work). 

Running `ip route` shows that the managed route added by ZeroTier has no metric, and so is defaulting to the highest priority. Manually changing this to a higher metric solves the problem.

I'm not sure what the best approach for this situation is, but interestingly it doesn't seem to be a problem on Android.

glimberg:
@novirium Well that's a weird one.  ZeroTier definitely should be setting a metric such that the ZeroTier route is a lower priority than the physical network.  What Linux distribution & version of the distro are you using?

And no, it wouldn't be a problem on Android.   Android & iOS are entirely different beasts.

novirium:
I'm not seeing any metric on the added route on both Arch Linux and Debian Stretch (it's probably worth a separate issue, but on the Stretch host the local subnet route is via a bridge, and ZeroTier is removing that route and replacing it entirely). Both are running ZeroTier One 1.2.8 .

For clarity, the local subnet is 10.0.3.0/24, and the zerotier network is 10.5.5.0/24. The gateway node (10.5.5.51) is connected to both, and I have the managed route `10.0.3.0/24 : 10.5.5.51` added in Zerotier Central.

[...]

PR #1455, opened 2021-09-29
allow to setup metric for created routes (linux only)

rjsocha:
This hack is adding support for creating routes with metrics.
This is not for inclusion. More to open the door for discussion (if any).

By default, routes are created with metric 100000 (zerotier-one -mMETRIC to change this, 0 - to disable metric manipulation).

What is solve for me?
I have a roaming laptop and in one of my networks (ZT is installed on the gateway) if the laptop is connected via wifi+lan all local traffic is routed via ZT interface.
This is the same problem described in this issue: https://github.com/zerotier/ZeroTierOne/issues/750


rjsocha:
Solved in #750

PR #1467, opened 2021-10-28
Fix docker entrypoint.sh startup check

nicksmith53:
Fixes a loop when restarting an existing ZeroTier container.


nicksmith53:
Fixed in f41372680d11c36c01e43e922c781845224dd62d

erikh:
sorry I apparently missed this ticket. thanks for doing this!

PR #1480, opened 2021-11-03
dockerfile: Use apt repository instead of dpkg directly

daurnimator:
In addition to being better practice, this helps make things architecture independant


someara:
Hey sorry we slept on this for so long.
Closing, as stale

PR #1530, opened 2022-01-05
Disable Salsa20 SSE when ZT_SALSA20_SSE=0

StephenCWills:
This allows us to force disable the `ZT_SALSA20_SSE` option even when `__WINDOWS__` or `__SSE__` is defined.

The 32-bit MinGW compiler for Windows doesn't support the compiler intrinsics used by the Salsa20 SSE implementation.

```
In file included from /home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.hpp:23,
                 from /home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.cpp:11:
/usr/lib/gcc/i686-w64-mingw32/9.3-win32/include/emmintrin.h: In constructor ‘_s20sseconsts::_s20sseconsts()’:
/usr/lib/gcc/i686-w64-mingw32/9.3-win32/include/emmintrin.h:1166:1: error: inlining failed in call to always_inline ‘__m128i _mm_slli_epi64(__m128i, int)’: target specific option mismatch
 1166 | _mm_slli_epi64 (__m128i __A, int __B)
      | ^~~~~~~~~~~~~~
/home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.cpp:60:28: note: called from here
   60 |   maskHi32 = _mm_slli_epi64(maskLo32, 32);
      |              ~~~~~~~~~~~~~~^~~~~~~~~~~~~~
In file included from /home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.hpp:23,
                 from /home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.cpp:11:
/usr/lib/gcc/i686-w64-mingw32/9.3-win32/include/emmintrin.h:1432:1: error: inlining failed in call to always_inline ‘__m128i _mm_shuffle_epi32(__m128i, int)’: target specific option mismatch
 1432 | _mm_shuffle_epi32 (__m128i __A, const int __mask)
      | ^~~~~~~~~~~~~~~~~
/home/runner/work/devilutionX/devilutionX/build/_deps/libzt-src/ext/ZeroTierOne/node/Salsa20.cpp:59:31: note: called from here
[...]

PR #1531, opened 2022-01-05
Disable Salsa20 SSE on MinGW 32-bit compiler

StephenCWills:
This fixes the error described in #1530 by disabling Salsa20 SSE on the 32-bit MinGW compiler.

It may be worth noting that both the 64-bit and 32-bit MinGW compilers define the `__MINGW32__` variable, but the 64-bit compiler also defines `__SSE2__` so this works as expected on both.

### DIFF
--- a/node/Salsa20.hpp
+++ b/node/Salsa20.hpp
@@ -15,7 +15,7 @@
 #include "Constants.hpp"
 #include "Utils.hpp"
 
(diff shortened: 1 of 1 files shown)


StephenCWills:
Er.. Hold on, I forgot to update the commit message.

glimberg:
Thanks!

PR #1534, opened 2022-01-07
Windows interface and route metrics

laduke:
5 is the lowest metrics an interface will get based
on Windows Automatic Metrics as far as I know.
https://docs.microsoft.com/en-us/troubleshoot/windows-server/networking/automatic-metric-for-ipv4-routes

If we set our metric to 3, there's a chance Windows will
use it for broadcast traffic, and things like game discovery
will work. If the user doesn't want that, they can leave the
zerotier network or manually set the metric on their
physical interface to < 3.

For routes with "via"'s we put the metric high,
in case it overlaps with a physical route to
the same subnet.

These could be plumbed through and set via the controller or
local.conf someday, istead of being hardcoded magic numbers.

### DIFF
--- a/osdep/ManagedRoute.cpp
+++ b/osdep/ManagedRoute.cpp
@@ -292,14 +292,17 @@ static void _routeCmd(const char *op,const InetAddress &target,const InetAddress
 static bool _winRoute(bool del,const NET_LUID &interfaceLuid,const NET_IFINDEX &interfaceIndex,const InetAddress &target,const InetAddress &via)
 {
 	MIB_IPFORWARD_ROW2 rtrow;
+	const int ROUTE_METRIC = 5000;
 	InitializeIpForwardEntry(&rtrow);
 	rtrow.InterfaceLuid.Value = interfaceLuid.Value;
 	rtrow.InterfaceIndex = interfaceIndex;
+	rtrow.Metric = -1;
 	if (target.ss_family == AF_INET) {
 		rtrow.DestinationPrefix.Prefix.si_family = AF_INET;
 		rtrow.DestinationPrefix.Prefix.Ipv4.sin_family = AF_INET;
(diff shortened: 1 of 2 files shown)


laduke:
Caveat for the **Interface** metric: 

There are a small amount of service providers out there that use zerotier as just a remote access into user's PC. They don't necessarily want the zerotier network to take over broadcast on the user's machine.

glimberg:
Yeah.  We're probably going to want a flag a la `allowDefault` to enable/disable this

laduke:
took a look at this again. It seems like the config needs to go in the local network settings ala allowManaged, like grant said above. Which is not local.conf nor the network config from the controller. 

```
cat ./networks.d/d5e04297a1123456.local.conf 
allowManaged=1
allowGlobal=0
allowDefault=0
allowDNS=0
```

something like

```
cat ./networks.d/d5e04297a1123456.local.conf 
allowManaged=1
allowGlobal=0
allowDefault=0
allowDNS=0
interfaceMetric=2
routeMetric=5000
```

Not every OS has interface metrics or route metrics, so their could be a little confusion. Changing those values could be added to a UI in the future. 
It's a lot of plumbing. I'm not sure if I'm going to do try it right now.

laduke:
🙅  I think integrating something like winipbroadcast is better.

someara:
Rebasing the dev branch into this to keep it from drifting too far

laduke:
Thanks! I'm gonna close this. I don't think we want to do it this way.
When we delete the branch, does the extra context in the PR go away?

laduke:
```
RegSetKeyValueA(tcpIpInterfaces, _netCfgInstanceId.c_str(), "InterfaceMetric", REG_DWORD, &metric, sizeof(metric));
```

PR #1535, opened 2022-01-07
Set a high metric on linux routes that have a via.

laduke:
See issue #750.

zerotier doesn't currently set a metric on routes. Linux
takes this to mean "0", the highest priority.

Every comment in the issue is about routing between zerotier
and lan and how they conflict.

This quick change could fix this problem for most people.

The subnet route for the zerotier network, the one with no
via, is still 0 in this patch. Just the "via" routes get
higher metrics.
If for some reason, you needed your via routes to have
higher priority, you could use a prefix work-around:

192.168.1.0/25 via 10.147.17.1
192.168.1.128/25 via 10.147.17.1

### DIFF
--- a/osdep/LinuxNetLink.cpp
+++ b/osdep/LinuxNetLink.cpp
@@ -25,6 +25,7 @@
 #ifndef IFNAMSIZ
 #define IFNAMSIZ 16
 #endif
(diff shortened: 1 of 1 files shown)


ihipop:
when will this feature released 
I didn't see any thing mention this in RELEASE NOTE

Issue #1551, opened 2022-01-27
Can't communicate sometimes when using Relay


Shawn8410:
My Enviroment:
One Node as Moon，Two Node as Peers。The Peers can't direct contact each other because of the damn security policy，So they must communicate through the Moon as Relay. 

Expecting:
The two Peers can communicate with each other.

Acutally Happening:
Sometimes they can contact each other, but after some hours later( Like 2 Hours，not so stable)，the communication will lost. And sometimes it will repaired a few minutes later, and other times will never recover until the peers restarted.

Because this bug can't be reproduced with regular, I thought it's a wrong configuration at first. But after I split the network node,  root node and the {network,root} node not join the network, this situation happened still. So I dived into code, and I think I find the reason of the bug.

My English is poor,  I try to describe it as clearly as I can.

After peers Joined a network, they first get their own com(CertificationOfMembership)，and   when the two peer try to communicate with each other, they first send a  MULITI_FRAME packet (ARP) to other peer, and because the other peer doesn't have its com, will responde a ERROR packet with  ERROR_NEED_MEMBERSHIP_CERTIFICATE, when the peer receive the ERROR packet, it will send its com by a NETWORK_CREDENTIALS packet to the other peer, and after the other peer accept the com, they can communicate in this network.  Like:
![image](https://user-images.githubusercontent.com/94963436/151286656-ad5095df-23a4-475a-bad0-9eed8e1a6605.png)

This is beatiful for building a secure private network access! 
But by the same time, Zerotier try to make a intranet direct access if possible, its a PUSH_DIRECT_PATHS packet send to each other after every trusted packet it received( Peer::received), In order to avoid flooding send, it will send at maximum every 120s by direct, 15s by relay.

Similarly, to avoid flooding,  peer will controll its receive frequency for some cotroll packet, such as NETWORK_CREDENTIALS, PUSH_DIRECT_PATHS(Peer::rateGate*). And there the bug is:

```
	/**
	 * Rate limit gate for VERB_NETWORK_CREDENTIALS
	 */
	inline bool rateGateCredentialsReceived(const int64_t now)
	{
		if ((now - _lastCredentialsReceived) <= ZT_PEER_CREDENTIALS_CUTOFF_TIME)
			++_credentialsCutoffCount;
		else _credentialsCutoffCount = 0;
		_lastCredentialsReceived = now;
		return (_directPathPushCutoffCount < ZT_PEER_CREDEITIALS_CUTOFF_LIMIT);
[...]

PR #1552, opened 2022-01-27
FIX: fix wrong flag usage of rateGateCredentialsReceived

Shawn8410:
#1551 

Fix for the occasionally disconnect of peers


joseph-henry:
Looks great. Thanks for finding this. Going to do a little testing before we merge this in.

Issue #1554, opened 2022-02-02
Managed  MTU doesn't get applied. 


laduke:
If you change a network's MTU, your interface still gets 2800, on Linux at least.  

Leaving/Joining doesn't change it. If you restart zerotier-one, you get the network's value, but if you leave/join, it goes back to 2800. 

Probably a netlink ordering thing.

laduke:
Actually, the fact that it's getting set to 2800 implies it's using the default instead of the network config

glimberg:
Hrmm, so for joining, it looks like zerotier is creating the interface with the default MTU before even receiving the network config?

glimberg:
Assuming the network config is coming in correct, it should be correct on `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP`

https://github.com/zerotier/ZeroTierOne/blob/269501eaa0c22bdc402e689b0d061325fb6ddbce/service/OneService.cpp#L2426

laduke:
https://github.com/zerotier/ZeroTierOne/blob/269501eaa0c22bdc402e689b0d061325fb6ddbce/node/Network.cpp#L587-L593

https://github.com/zerotier/ZeroTierOne/blob/269501eaa0c22bdc402e689b0d061325fb6ddbce/node/Network.cpp#L1390-L1400

-> on join,  mtu is 2800

🐇

qt1:
Hi,

For me Setting mtu with local.config does not seem to work. Not sure if this is a different problem. (or I am not setting the mtu in the correct way)

    sudo zerotier-cli info -j

gives

````
{
 "address": "1234567890",
 "clock": 1647550455798,
 "config": {
  "physical": {
   "72.28.0.0/16": {
    "mtu": 1200
   }
  },
  "settings": {
   "allowTcpFallbackRelay": true,
   "listeningOn": [
    "192.168.122.205/9993",
    "192.168.122.205/29994",
    "192.168.122.205/48689"
   ],
   "portMappingEnabled": true,
   "primaryPort": 9993,
   ...
````

But the actual mtu is 2800

Changing the mtu with `ip link set` does work but it is no persistent.

glimberg:
@qt1 MTU is not settable via `local.conf`.  It's a network level option controlled by the network controller.

qt1:
@glimberg ,

How should I set it the mtu?
I do not see any dhcp option in the UI and the link is controlled by ZT , but I didn't find a way to set it there.
[...]

PR #1559, opened 2022-02-08
potential fix for 2.6.x kernels

glimberg:
detect if the kernel is super old.  if so, do not execute the post IF_UP commands on the interface

@laduke Also made a slight MTU change.  Can you see if this also fixes https://github.com/zerotier/ZeroTierOne/issues/1554?


adamierymenko:
Testing merge, will revert if there are issues. Seeing if it fixes things on Ye Olde Tyme Linux Distributions like CentOS 6.

PR #1569, opened 2022-02-22
Rate gate ECHO per Path instead of per Peer

joseph-henry:
### Summary

In multipath scenarios user traffic is used to judge the aliveness of a path. If the user traffic is too infrequent to establish aliveness for a given time window (say `250 ms`), the bonding layer will send extra `ECHO`s at a maximum rate of `failoverInterval / 4` (or `62.5 ms`) *per path*. This patch relaxes the rate-limiting of `ECHO`s significantly in order to prevent a non-multipath node from dropping `ECHO`s causing multipath nodes to erroneously judge paths to that node to be dead.

## Details

This patch decreases the rate limiting from `1000 ms` by a factor of `20` to `50 ms` and rate limits `ECHO`s per `Path` instead of per `Peer`. This allows rate limiting to scale with the number of established paths to a peer.

As a result, if all `64` path slots are used a total of `64 x 16 = 1024` `ECHO`s per second will be allowed in the most aggressive case where `failoverInterval` is set to `250 ms`.

### DIFF
--- a/node/Constants.hpp
+++ b/node/Constants.hpp
@@ -571,13 +571,13 @@
  * Anything below this value gets into thrashing territory since we divide
  * this value by ZT_BOND_ECHOS_PER_FAILOVER_INTERVAL to send ECHOs often.
  */
-#define ZT_BOND_FAILOVER_MIN_INTERVAL 250
+#define ZT_BOND_FAILOVER_MIN_INTERVAL 500
 
 /**
  * How many times per failover interval that an ECHO is sent. This should be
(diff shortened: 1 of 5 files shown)


joseph-henry:
More thoughts on this:

Setting the minimum `failoverInterval` to `500 ms` and reducing probe period from `failoverInterval / 4` to `failoverInterval / 3` would result in a max of `6 ECHOs per second per path` which is `384 ECHOs per second per peer` if all 64 path slots are filled.

adamierymenko (approved):

PR #1596, opened 2022-03-18
Fix dockerfile & entrypoint:

erikh:
- Resolve issue with join not being checked properly for success without
  using external tools
- Resolve issue where initial boot was not being checked properly
- Now output errors when zerotier fails to start

closes #1581

cc @altano for inspiration for this patch

Signed-off-by: Erik Hollensbe <git@hollensbe.org>

PR #1603, opened 2022-03-22
Fix for missing dependencies in Dockerfile.release

MaanuelMM:
When building my own Docker image of ZeroTier with the latest version 1.8.6 using the `docker build -f Dockerfile.release -t zerotier/zerotier:latest --build-arg VERSION=1.8.6 .` command, I noticed it failed due to missing dependencies related to SSL that the _Dockerfile.release_ did not take into account:

```console
dpkg: dependency problems prevent configuration of zerotier-one:
 zerotier-one depends on libssl1.1 (>= 1.1.1); however:
  Package libssl1.1 is not installed.
 zerotier-one depends on openssl; however:
  Package openssl is not installed.

dpkg: error processing package zerotier-one (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 zerotier-one
```

To fix this error, I've decided to make use of the _APT_ package manager instead of the _DPKG_ tool to install the _DEB_ package, thus resolving the dependencies automatically and successfully creating the Docker image.

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -13,7 +13,10 @@ FROM debian:buster
 
 COPY --from=stage zerotier-one.deb .
 
-RUN dpkg -i zerotier-one.deb && rm -f zerotier-one.deb
+RUN apt-get update -qq \
+  && apt-get install -y ./zerotier-one.deb \
+  && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


erikh:
This is covered by #1596. I appreciate the patch but this is also not the direction we want to head.

PR #1614, opened 2022-04-02
Add loongarch64 support

fuowang:
Add  loongarch64 support, the new architecture launched by China's Loongson.


joseph-henry:
@fuowang. Thanks. Given that we don't have any access to this architecture to test, I have a few questions:

- Have you verified that ZeroTier does compile and run?
- Which chips have you tested? (3A5000)?
- How is ZeroTier's performance on this chip?
- Which *commercial* products/devices can you point to that currently use this architecture?

fuowang:
Yes, I have already compiled (no drone build) and use it for about half a month, everything is good. I use it on the Loongson-3A5000 chip, theoretically, the 3C5000 also supports.  

Linux release: Kylin Linux Advanced Server release V10 (Tercel)  

In this architecture, `zerotier-one -d` runs stable, from March 24, it has never been interrupted, and the network connection is good:  


Loongson 3A5000: <https://loongson.cn/product/show?id=10>
Loongson 3C5000: <https://loongson.cn/product/show?id=15>
Loongson 3C5000L: <https://loongson.cn/product/show?id=11>

There are currently available commercial products/devices with this architecture:
Motherboard / development board:  
<https://item.jd.com/100033374408.html>  
<https://item.jd.com/100035752766.html>  
Laptop:  
<https://item.jd.com/100016595171.html>  
<https://item.jd.com/100032993180.html>
<https://item.jd.com/100028944178.html>  
Desktop:  
<https://item.jd.com/100023656622.html>  
<https://item.jd.com/100015834993.html>  
<https://item.jd.com/100024659270.html>

joseph-henry:
Thanks. We'll accept this if the drone config is removed. Since we don't possess any loongarch64 hardware we cannot really test it right now anyway. Additionally since it is essentially just rebranded MIPS we can assume it works. If in the future there is more demand for support we can revisit adding this to our CI.

fuowang:
Thanks a lot, the drone configuration has been removed. In fact, loongarch64 is not just rebranded MIPS, it is redesigned and contains many new instructions. Also, zerotier-one runs very stable on my loongarch64 machine.

Issue #1587, opened 2022-03-13
Bogus characters appearing in "ifname" field


dhbaird:
Overview: I do a cable-unplug to see if multipath fails over. What I found is some corrupted output in API requests.

- Expected: Correct interface names to be shown in JSON API.

- What is actually happening?
```
curl -s -H @<(echo -n "X-ZT1-Auth: "; cat /var/lib/zerotier-one/authtoken.secret) \
    http://localhost:9993/bond/show/XXXXXXXXXX
# >>>
{
 "bondingPolicy": "balance-aware",
 "downDelay": 0,
 "failoverInterval": 5000,
 "isHealthy": false,
 "links": [
  {
   "ifname": "wlan0",
   "path": "XX.XXX.XX.XX/57180"
  },
  {
   "ifname": "wlan0",
   "path": "XX.XXX.XX.XX/57180"
  },
  {
   "ifname": "\u0002",           <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/9993"
  },
  {
   "ifname": "\u0001",           <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/57179"
  },
  {
   "ifname": "\u0001",           <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/57180"
  },
  {
   "ifname": "/7.74.0",          <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/57179"
  },
  {
   "ifname": "/7.74.0",          <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/57180"
  },
  {
   "ifname": "/7.74.0",          <--- expect "eth0"
   "path": "XXX.XXX.X.XXX/9993"
  },
  {
   "ifname": "wlan0",
   "path": "XX.XXX.XX.XX/57180"
  }
 ],
 "numAliveLinks": 2,
 "numTotalLinks": 9,
 "upDelay": 0
}
```

- To reproduce, (1) set local.conf from below, and then (2) unplug eth0 and wait for failover. (Note: this failover interval takes a **long** time, several minutes instead of mere seconds, which is another issue Also trying "active-backup" mode results in a crash loop with "zerotier-one.service: Main process exited, code=killed, status=11/SEGV". I have a feeling I still may not be using multipath as intended).
```
cat > /var/lib/zerotier-one/local.conf << EOF
{
  "settings": {
    "defaultBondingPolicy": "myPolicy",
    "policies": {
      "myPolicy": {
[...]

PR #1658, opened 2022-05-05
Fix SEGV caused by format string on 32-bit platforms / armv7l, reported in #1587

dhbaird:
Fixes SEGV in #1587. Happy to re-submit if any changes are desired. Thanks!


dhbaird:
Note: I just force pushed over this to convert a spaces indent into tabs.

Issue #915, opened 2019-02-05
settings/primaryPort not fully respected


darkain:
**Describe the bug**
Relates to the issues with: https://github.com/zerotier/ZeroTierOne/issues/779

If a connection cannot be established on the primary port (usually 9993), then ZeroTier will pick an arbitrary port to try and use instead. This defeats the entire purpose of having a primary port configuration. I have specific firewall rules in place for 9993 for a combination of security and to deal with issue #779. However, at some point ZT added support to "try other ports" that are outside of the configuration, which means the firewall rules in place for #779 no longer work, which means route flapping returned, broken connections, and excessively high CPU usage again. I also tried disabling port mapping to see if that was the issue, but this didn't resolve the issue either. There needs to be a way to force ZT to use one port and one port ONLY.

 - OS: FreeBSD 11.2
 - ZeroTier Version: 1.2.12
 - Hardware: Any (several tested)

cferrey:
Just curious: does anyone know if this was addressed through a recent update?  I was getting this flapping behavior pretty regularly in a bridged-LANs-over-WAN setup, but it stopped about 2-3 weeks ago and everything's been stable since then.  I didn't change anything, so am wondering if this got patched.

laduke:
[source](https://github.com/zerotier/ZeroTierOne/blob/99c3f02343a7cb704dbde0b2e324451090875d93/service/OneService.cpp#L711)

Seems like a small change... but how to specify it?

Should just using `-p` on the commmand line turn off secondary port and upnp port? That's probably not what you want most of the time. Most users are behind NAT.


`'secondaryPortEnabled': true|false`   to go along with portMappingEnabled? 

or how about

`'maxBindPorts': 1-32` (default 3) heh

KBrownConsulting:
I think I'd vote for something along the lines of option 2... Clear individual options for controlling whether UPnP & the secondary port are enabled or disabled.

darkain:
For firewall security reasons, I need the ability to fully control all incoming/opened ports on my server/router. This is ESPECIALLY true when talking about security audit compliance, such as PCI-DSS. Every open port must be justified to an external audit company, and with ZT arbitrarily opening additional ports if there are ANY issues at all with the defualt 9993 just causes havoc for these security audits (as well as the route flapping issue causing 100% CPU usage and dropped packets)

joseph-henry:
@darkain Thanks, we've added the ability to turn off any non-primary port assignments via:

[...]

PR #1662, opened 2022-05-09
Improve ephemeral port restriction handling

joseph-henry:
This change makes it easier to control when ephemeral ports can be used via `local.conf` (See issue https://github.com/zerotier/ZeroTierOne/issues/915)

New behavior:

- The `allowSecondaryPort` option in `local.conf` is replaced by `allowEphemeralPorts`
- `allowEphemeralPorts` is a toggle that controls all non-9993 ports (e.g. `secondaryPort` to get around buggy NATs and `tertieraryPort` for uPnP/NAT-PMP)

Why this change is needed:

- There does not exist a single option switch to control all non-9993 ports. One would have to set `allowSecondaryPort:false` and `allowPortMapping:false`.

Open questions:

- Should this change *replace* `allowSecondaryPort` causing a breaking change or should we leave that switch for backwards compatibility (introducing possibly confusing redundancy)?
- Is there a situation where people would want to disable the secondary port but leave the tertiary port?

### DIFF
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -750,7 +750,7 @@ public:
 	PhySocket *_localControlSocket6;
 	bool _updateAutoApply;
 	bool _allowTcpFallbackRelay;
-	bool _allowSecondaryPort;
+	bool _allowEphemeralPorts;
 
 	unsigned int _primaryPort;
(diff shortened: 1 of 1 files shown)


joseph-henry:
Closing due to: Not important any longer

Issue #1615, opened 2022-04-03
[Feature Request] Launch Script/Tool on network up/down


erikh:
I'd like it if zerotier-one could launch a script on network up/down. I don't think this can (adequately) be resolved by external tooling.

This way I could prime resolvers for DNS, similar to what one might do with a DHCP client.

ftrojahn:
+1

joseph-henry:
Hmm. This tickles my security spidey sense. Could one just monitor network status changes via something like `zerotier-cli listnetworks -j | jq .[].status`

If we did implement something like this would just invoking a `hook.sh` and passing a network ID as an argument work for you?

erikh:
yep

------- Original Message -------

joseph-henry:
Ok, there's something. Let me know what you think.

PR #1669, opened 2022-05-17
Add system script hook for network events

joseph-henry:
This patch adds the ability to call a user-specified script upon various network events (See #1615). I haven't yet considered *all* of the security implications of this so let's not merge until we've thought hard about this. 

Example `/var/lib/zerotier-one/hook.sh`:

```
#!/bin/sh

network_up()
{
	echo "network_up $1" >> log.txt
}

network_update()
{
	echo "network_update $1" >> log.txt
}

network_down()
{
	echo "network_down $1" >> log.txt
}

"$@"
```

Example `local.conf`:

```
{
	"settings":
	{
		"eventHookScriptPath": "/var/lib/zerotier-one/hook.sh"
	}
}
```

Example `ZT_TRACE` output:

```
Running network event hook script (/var/lib/zerotier-one/hook.sh)
```

### DIFF
--- a/osdep/OSUtils.cpp
+++ b/osdep/OSUtils.cpp
@@ -11,34 +11,34 @@
  */
 /****/
 
-#include <stdio.h>
-#include <string.h>
-#include <stdlib.h>
-#include <stdarg.h>
-#include <sys/stat.h>
-#include <stdlib.h>
-
 #include "../node/Constants.hpp"
 #include "../node/Utils.hpp"
 
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+
 #ifdef __UNIX_LIKE__
-#include <unistd.h>
+#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
-#include <sys/types.h>
+#include <netdb.h>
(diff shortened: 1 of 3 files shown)


osdep/OSUtils.cpp
```diff
@@ -547,6 +547,43 @@ std::string OSUtils::jsonBinFromHex(const nlohmann::json &jv)
 	return std::string();
 }
 
+void OSUtils::_hookCmd(const char* scriptPath, const uint64_t nwid, ZT_VirtualNetworkConfigOperation op)
+{
+	if (! scriptPath || ! strlen(scriptPath) || nwid == 0) {
+		fprintf(stderr, "no script path specified\n");
+		return;
+	}
+	long p = (long)fork();
```
erikh, line 556: probably want to scrub environ here

erikh: also need a SIGCHLD handler here I think?

erikh (commented):
I'm probably wrong about all this, but this is what I saw as issues

joseph-henry:
Good questions, I'm using the same pattern we use to run other system commands. Didn't want to deviate too much.

erikh:
fair enough

joseph-henry:
Closing due to: Not important any longer

PR #1680, opened 2022-05-23
Default route mac

laduke:


dmkjr:
I, too, have noticed duplication on networks.  This would be a nice add.

laduke:
I"m not sure this should go in as it was. Wasn't it marked as a "discussion" and not a "pr"?

PR #1683, opened 2022-05-30
Add redhat/ubi:8 dockerfile for s390x support

srajmane:
Add redhat/ubi:8 dockerfile for s390x support


srajmane:
@glimberg @joseph-henry Could you please have a look?

PR #1701, opened 2022-06-19
fix sql

tossp:
bind message supplies 17 parameters, but prepared statement "" requires 16


erikh:
ahem
WELP

glimberg:
already fixed elsewhere

glimberg:
thought it was, but nope it wasn't.  Good catch

PR #1704, opened 2022-06-21
Synology Docker: update entrypoint.sh

deekue:
- propagate TERM/QUIT/INT signals
- add some basic logging
- check for unbound variables
- update "route helper"
   - run as subshell, exit if zerotier-one is unavailable so pod can be restarted
   - only call `zerotier-cli` once, avoids race conditions
   - only add default routes if allowDefault is enabled for that network
   - add some more error handling
   - sleep after all networks are processed

### DIFF
--- a/pkg/synology/dsm7-docker/Dockerfile
+++ b/pkg/synology/dsm7-docker/Dockerfile
@@ -2,13 +2,12 @@
 
 FROM alpine:latest as builder
 
-RUN apk add --no-cache rust cargo
-RUN apk add  openssl-dev
-
-RUN apk add --update alpine-sdk linux-headers \
+WORKDIR /src
+RUN apk add --no-cache rust cargo \
+  && apk add  openssl-dev \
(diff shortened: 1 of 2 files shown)


joseph-henry:
Thanks, this looks like really good work. Testing this now. The only problem that I see is:

At the beginning,

```
if ! NETWORK_LIST="$(zerotier-cli -j listnetworks)"
then
  echo "Route helper: $NETWORK_LIST" >&2
  exit 1
fi
```

If ZT takes longer than 15s to start up (low-powered CPU devices could when generating an initial node ID) this script will exit and never be called again. This would result in routes failing to be added. I might lengthen that initial wait interval.

erikh:
poll it

deekue:
switched to polling and made the wait and sleep interval configurable via Docker env vars.  
Defaults are 15s sleep and 90s total wait.

joseph-henry:
I've tested this and verified that it works. Thank you for this excellent work. I'll merge this and push new builds momentarily.

PR #1706, opened 2022-06-21
fix erase network failed

tossp:
`Error deleting network: [json.exception.type_error.302] type must be string, but is null`

https://github.com/zerotier/ZeroTierOne/blob/dev/controller/PostgreSQL.cpp#L1481-L1491


adamierymenko:
This is actually obsolescent code, so closing for now.

PR #1736, opened 2022-08-10
Fix service installation MAX_PATH bug on Windows

visuve:
- On newer Windows, a path might hold up to 32,767 characters
  - https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
- The previous installation did not also check for ERROR_INSUFFICIENT_BUFFER which could lead to an ill formatted path


visuve:
Also note that the whole class uses ANSI versions of Win32 API functions. UTF-16 wide strings would be preferred, i.e. functions with W postfix.

PR #1750, opened 2022-09-07
Prevent re-armoring of packets when in multipath broadcast mode

joseph-henry:
Add a function `isEncrypted` to `Packet` that checks for the presence of set cipher flags. Then check this state in `Switch` before armoring a packet. Packets will only be armored if neither of the cipher flags are set. Normally there would never be a situation where this is needed but in the multipath `broadcast` scenario a packet will be armored and sent on multiple interfaces. It only needs to be armored once.

### DIFF
--- a/node/Packet.hpp
+++ b/node/Packet.hpp
@@ -1250,6 +1250,14 @@ public:
 	}
 
 	/**
+	 * @return Whether this packet is currently encrypted
+	 */
+	inline bool isEncrypted() const
+	{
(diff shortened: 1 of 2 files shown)


joseph-henry:
```
 ____________________
< Good morning, Adam >
 --------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

adamierymenko:
This looks like it should be fine, but please add a check for unencrypted packets in the receive path and test. Should not be possible but check.

Issue #1734, opened 2022-08-09
Poor Multipath TCP throughput


UnmovableCheddar:
I have Many-to-One multipath configured with 2 WAN links on site A and 1 WAN link on site B.

Both of site A links are symmetric 100 Mbps.
Site B link is 1 Gbps.

Site B is configured as a default gateway for all Site A oubound connections, and the only NAT is ZT > WAN. There is no NAT for the tunnel.

Latency from Site A to Site B is ~40ms and both are using ZT version 1.10.1.

With balance-xor or balance-aware on both sides, I'm able to get 200 Mbps with iperf3 UDP:
![](https://i.imgur.com/WazbLAj.png)

However, iperf3 TCP (10 parallel connections) with same settings is extremely slow:
![](https://i.imgur.com/wXndacO.png)

Testing a single stream wget download is better, but still unable to achieve even a single link speed's throughput:
![](https://i.imgur.com/z82qqZ1.png)

Site A local.conf:
```
{
    "physical": {
        "192.168.0.0/16": {
            "blacklist": true
        }
    },
    "settings": {
        "allowSecondaryPort": false,
        "interfacePrefixBlacklist": [
            "wg",
            "zt"
        ],
        "allowTcpFallbackRelay": false,
        "portMappingEnabled": false,
        "defaultBondingPolicy": "custom-balance-xor",
        "policies": {
            "custom-balance-aware": {
                "basePolicy": "balance-aware",
                "rebalanceStrategy": "aggressive",
                "balancePolicy": "flow-dynamic",
                "links": {
                    "igb1": {},
                    "igb2": {}
                }
            },
            "custom-balance-xor": {
                "basePolicy": "balance-xor",
                "rebalanceStrategy": "aggressive",
                "links": {
                    "igb1": {},
                    "igb2": {}
                }
            }
        }
    }
}
```

Site B local.conf:
```
{
    "physical": {},
    "settings": {
        "allowSecondaryPort": false,
        "interfacePrefixBlacklist": [
            "wg"
        ],
        "portMappingEnabled": false,
        "allowTcpFallbackRelay": false,
        "defaultBondingPolicy": "custom-balance-xor",
        "policies": {
            "custom-balance-aware": {
                "basePolicy": "balance-aware",
                "balancePolicy": "flow-dynamic",
                "rebalanceStrategy": "aggressive"
            },
            "custom-balance-xor": {
                "basePolicy": "balance-xor",
[...]

PR #1752, opened 2022-09-13
Fix custom policy bond parameters

grapexy:
This should fix flow assignment issues with custom policies as highlighted in https://github.com/zerotier/ZeroTierOne/issues/1734#issuecomment-1221619798


joseph-henry:
Thanks for this. Your efforts aren't going unnoticed. A long while back certain features were deliberately disabled because of bugginess and the docs weren't updated accordingly. It was planned to re-introduce them but work in other areas was needed first in order to stabilize what `balance-aware` was to be built atop. Once I'm done with a few other issues I'll see how I can address your other tickets and possibly merge this.

grapexy:
Thanks for building the most understandable and tweakable multipathing solution I've dealt with! I can't wait until `balance-aware` is fully functional. Even if buggy, it's still the best open source solution right now.

PR #1765, opened 2022-10-25
Expose surface addresses in info json

laduke:
This just adds a node's "surface addresses" to the output of `zerotier-cli info -j` 

----

Surface Addresses are the addresses that
the roots report back to you.

This is helpful for troubleshooting.

If you're behind NAT, the source port is different than what zerotier is bound to.
If the list of surface address ports is larger than the list of bound addresses, you are probably behind symmetric NAT.

Anyways this can be added to later with a more simple "easy" or "hard" nat computed message somewhere.

### DIFF
--- a/node/Node.hpp
+++ b/node/Node.hpp
@@ -35,6 +35,7 @@
 #include "NetworkController.hpp"
 #include "Hashtable.hpp"
 #include "Bond.hpp"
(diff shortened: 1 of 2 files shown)


joseph-henry:
Been testing it for an hour on my machine with both a memory leak checker and address sanitizer. Everything checks out. Well done.

Issue #1794, opened 2022-11-27
About Compilation Errors


TcDhlPro:
---
Directory: `ZeroTierOne/attic/world`
My Actions: `./build.sh`
Error Content: ↓↓↓↓↓↓
---
![image](https://user-images.githubusercontent.com/100897011/204115087-1b7f82cd-c879-4d01-b284-45c31a7a5a7a.png)

---
System version: `Debian-10`
ldd --version: `ldd (Debian GLIBC 2.28-10) 2.28`
gcc -v: `gcc version 8.3.0 (Debian 8.3.0-6)`
---
How to solve this error?
Can I ignore it?
---

glimberg:
These are compiler warnings.  Not errors.  They're safe to ignore.

TcDhlPro:
> These are compiler warnings. Not errors. They're safe to ignore.

ok~ Thank you for your reply

PR #1798, opened 2022-11-29
fix warning: 'readdir_r' is deprecated: readdir_r is deprecated; use …

bostick:
…readdir instead

### DIFF
--- a/osdep/OSUtils.cpp
+++ b/osdep/OSUtils.cpp
@@ -101,15 +101,12 @@ std::vector<std::string> OSUtils::listDirectory(const char *path,bool includeDir
 		FindClose(hFind);
 	}
 #else
-	struct dirent de;
 	struct dirent *dptr;
 	DIR *d = opendir(path);
 	if (!d)
 		return r;
-	dptr = (struct dirent *)0;
 	for(;;) {
-		if (readdir_r(d,&de,&dptr))
-			break;
+		dptr = readdir(d);
 		if (dptr) {
 			if ((strcmp(dptr->d_name,"."))&&(strcmp(dptr->d_name,".."))&&((dptr->d_type != DT_DIR)||(includeDirectories)))
 				r.push_back(std::string(dptr->d_name));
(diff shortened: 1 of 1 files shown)


glimberg:
Not sure that this should be changed. `readdir` is not thread safe.  There's no external synchronization to prevent multiple threads reading the same thing at the same time.

bostick:
I think there are a few approaches of where to go, ranked in terms of amenability for me:

[1] https://www.gnu.org/software/libc/manual/html_node/Reading_002fClosing-Directory.html


## switch to readdir, no mutex

The page [1] gives multiple reasons why to use readdir and talks about being safe to read from different dirstreams.

In all occurences of `readdir_r` in OSUtils.cpp, the dirstream is stack-allocated where multiple threads would not be reading from same dirstream.

So there is probably no need for external synchronization and simple replacement with `readdir` can be done.


## switch to readdir, with mutex

If we do want to add external synchronization, adding a static mutex in OSUtils.cpp from `<mutex>` around `readdir` is easy:
```
rMutex.lock();
readdir();
rMutex.unlock();
```

This is the recommendation from [1]:

    In the rare case that multiple threads need to read from the same dirstream, it is still better to use readdir and external synchronization.

But again, all the dirstreams are different and external synchronization is probably not needed for cases in OSUtils.cpp.


## silence the warning

```
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```


## Ignore the warning

I would like to fix the warnings that occur when building ZeroTierOne, and simply ignoring the warning would be my least preferred path.

There is value in having a clean build, for both internal users and external users.

There was an issue from an external user complaining about the warning:
https://github.com/zerotier/ZeroTierOne/issues/1794

This would make external users happy.

adamierymenko:
For this one I am just going to pragma disable the warning. Linux technically deprecates this function but it still exists on every target distro we know, and other OSes prefer it. Don't think this needs to be fixed in the C++ code base.

PR #1803, opened 2022-12-02
Add termux build support: android-aarch64

JingMatrix:
For compatible issue, we use the /usr/var directory instead of the /var directory.

This diretory change might need discussions, please tell me your ideas.

### DIFF
--- a/make-linux.mk
+++ b/make-linux.mk
@@ -12,7 +12,11 @@ endif
 INCLUDES?=-Izeroidc/target -isystem ext
 DEFS?=
 LDLIBS?=
+ifdef TERMUX_VERSION
+DESTDIR?=$(PREFIX)/..
+else
 DESTDIR?=
+endif
 
 include objects.mk
 ONE_OBJS+=osdep/LinuxEthernetTap.o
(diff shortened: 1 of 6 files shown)


glimberg:
The directory change is not going to work for all Linux environments. There are far too many existing clients out in the wild to switch the directory out from under them on an upgrade.  If it's a requirement of Termux, you'll need to find a way to limit that directory change only to Termux builds.  

Additionally, if it's Termux only issue, you can specify the home directiory when starting ZeroTier.  `zerotier-one -d /path/to/custom/directory` will use that directory as the ZeroTier data folder.

JingMatrix:
OK, I will change it in the makefile

JingMatrix:
Now it should be Okay, directory changes are reverted, please consider merge.

daoudeddy:
I think this port of ZeroTier to Termux is doable, but your PR need more work.

1- instead of creating symbolic link of `$PRIFIX/usr/var` on `$PRIFIX/../var` all you need to do is change the dir to `$PREFIX/var` with a condition that checks if it is a Termux env in the make-linux.mk

2- as far as i know Termux doesn’t support `/sbin` since the device isn’t rooted by default, so the other thing you need to change is also replacing `$PREFIX/usr/sbin` with `$PREFIX/bin`

And everything should be fine

JingMatrix:
For the first point, I intended to make less changes to the original code when possible.
Otherwise, I should repeat the orginal installation code.

For the second point, it is impossible to run `zerotier-one` without root.
I agree that it could be a small problem since `/sbin` by default is not in the PATH.

daoudeddy:
> For the first point, I intended to make less changes to the original code when possible. Otherwise, I should repeat the orginal installation code.
> 
> For the second point, it is impossible to run `zerotier-one` without root. I agree that it could be a small problem since `/sbin` by default is not in the PATH.

Adding more code isn’t a bad idea, it will make thing clean and clear for future modifications.
There are some common commands you can use for both instead of duplicating.
 
 Also it is a good thing to add what dependencies you need to be able to build this on Termux.

PR #1820, opened 2022-12-14
Low Bandwidth Mode (for IoT)

joseph-henry:
This timer adjustment patch reduces ZeroTier's ambient network data usage by doing the following:

- Reduce frequency at which full `HELLO`s are sent to peers to `1/2` of original
- Reduce frequency at which full `HELLO`s are sent to upstream peers to `1/16`th of original
- Reduce frequency at which network configs are requested to `1/64`th of original
- Network config updates are no longer replicated to peers
- Will not send HELLOs out on all available local sockets (only one now.)
- Reduce frequency at which direct paths are pushed to peers to `1/16`th of original
- Will no longer push additional physical addresses if one is already known and workable
- Reduce frequency of multicast gather to `1/3`rd of original

Currently this new mode can be enabled by adding the following to `local.conf`

```
{
  "settings":
  {
    "lowBandwidthMode":true
  }
}
```

The side effects of this mode will be a ZeroTier that is slower to respond to network changes but should otherwise still act as expected. This works on my machines and has been tested on at least one other machine but should still be considered experimental. Since it must be manually enabled I think it is safe to include in the next release.

A future revision of this feature may be toggle-able via a `zerotier-cli` command.


glimberg:
Looks good to me on first look.  We should consider enabling it by default on mobile platforms once it's been put through its paces a bit more.

joseph-henry:
There's still a couple more items I want to add over time relating to how upstream peers are contacted (and why) but I think this is good enough as is for an initial release. I'd definitely want to get some feedback from people living with this mode on for a while before I'd be comfortable making it default anywhere. But yes, it could help on mobile.

I think the average user might find this mode annoying. But someone trying to SSH into a box once every few weeks likely won't notice the delays.

laduke:
Awesome!

People will ask for bandwidth numbers, if you happen to have them offhand from working on this. 
Stock zerotier idling is ~1GB / month IIRC

laduke:
Which of the list items is what keeps NAT hole-punch mappings alive?

someara:
Could we make this configurable via Central?

joseph-henry:
> Could we make this configurable via Central?

Potentially in a future version of this feature, yes.

joseph-henry:
@laduke To answer your question, the path heartbeats (empty `HELLO`s) have not been adjusted since I found their timings to be exactly tuned to appropriate minimum UDP keep alive requirements with minimal redundancy. I thought about stretching it a bit but the margins were so thin it might cause issues with minimal benefit. However full `HELLO`s have been reduced. So there should be no new NAT issues with regards to keep alive. If you're curious you can see `#define ZT_PATH_HEARTBEAT_PERIOD 14000`. It's set this way I believe so that two are sent ever `30s` which seems like a common UDP NAT timeout value, so we will likely send two of these every timeout period with `2s` of slack each time period which is I assume is there to allow for a little variance in latency.

joseph-henry:
@adamierymenko I believe this is ready for your eyes

PR #1825, opened 2022-12-20
enable ICMPv4 in windows firewall

laduke:
hey, for discussion

# motivation

Windows blocks ping by default. It allows every other kind of traffic though... 

That means people try ZeroTier, test with ping, and it doesn't work. 
Then they give up and try a different product, or they ask us and we give them instructions. The instructions are awkward for non-technical users. 

# implementation 
If `allowManaged` is allowed, poke the firewall for our specific interface and address. Don't touch the rules for the user's other interfaces or addresses. 

shell out to powershell:
`New-NetFirewallRule -DisplayName zerotier-icmpv4-d5e04297a1111111172.22.208.215 -InterfaceAlias 'ZeroTier One [d5e04297a1111111]' -Protocol ICMPv4 -Action Allow -LocalAddress 172.22.208.215"`

The DisplayName is used to delete entries when they aren't needed anymore

`Remove-NetFirewallRule -DisplayName zerotier-icmpv4-d5e04297a1111111*` remove all the entries related to d5e04297a1111111

`Remove-NetFirewallRule -DisplayName zerotier-icmpv4-d5e04297a1111111172.22.208.215` removes the entry for just one ip address (172.22.208.215)

# problems
- [x] ~~seems to take half a second to run these commands. I'm not sure if that blocks other important zerotier threads or not.~~ fixed this
- [x] ~~not sure how to check for old windows versions that don't have these commands, if applicable~~. windows 7 doesn't have the NetFirewallRule commands, but we don't run on Windows 7.
 

# alternatives
there is a c++ api to do this stuff, but I can't find api docs. There are [examples](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ics/c-adding-a-per-interface-rule) but they don't compile for me. There are also deprecation warnings for some of the functions they use.
It's beyond my c++ ability to write this in a time efficient way. If someone else wants to try that'd be awesome.

# todo 
- [x] ipv6, _should_ just be a simple copy paste of the ipv4 stuff

### DIFF
--- a/osdep/WinFWHelper.cpp
+++ b/osdep/WinFWHelper.cpp
@@ -1,0 +1,172 @@
+#include "WinFWHelper.hpp"
+
+
+namespace ZeroTier {
+
+
+
+void ZeroTier::WinFWHelper::newICMPRule(const InetAddress& ip, uint64_t nwid)
+{
+	char nwString[32] = { 0 };
+	char ipbuf[64];
+
+	sprintf(nwString, "%.16llx", nwid);
+	std::string nwString2 = { nwString };
+	
+	ip.toString(ipbuf);
+
+	if (ip.isV4()) {
+		WinFWHelper::newICMPv4Rule(ipbuf, nwid);
+	}
+	else {
+		WinFWHelper::newICMPv6Rule(ipbuf, nwid);
+	}
+}
+
+void ZeroTier::WinFWHelper::removeICMPRule(const InetAddress& ip, uint64_t nwid)
+{
+	char nwString[32] = { 0 };
+	char ipbuf[64];
+
+	sprintf(nwString, "%.16llx", nwid);
+	std::string nwString2 = { nwString };
+
+	ip.toString(ipbuf);
+
+	if (ip.isV4()) {
+		WinFWHelper::removeICMPv4Rule(ipbuf, nwid);
+	}
+	else {
+		WinFWHelper::removeICMPv6Rule(ipbuf, nwid);
+	}
+}
+
+
+void WinFWHelper::newICMPv4Rule(std::string address, uint64_t nwid)
+{
+	// allows icmp, scoped to a specific ip address and interface name
+
+	char nwString[32] = { 0 };
+	sprintf(nwString, "%.16llx", nwid);
+	std::string nwString2 = { nwString };
+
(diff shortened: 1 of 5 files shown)


glimberg:
I like it

osdep/WindowsEthernetTap.cpp
```diff
@@ -1250,6 +1256,88 @@ void WindowsEthernetTap::_setRegistryIPv4Value(const char *regKey,const std::vec
...
+		fprintf(stderr, cmd.c_str());
+
+		STARTUPINFOA startupInfo;
+		PROCESS_INFORMATION processInfo;
+		startupInfo.cb = sizeof(startupInfo);
+		memset(&startupInfo, 0, sizeof(STARTUPINFOA));
+		memset(&processInfo, 0, sizeof(PROCESS_INFORMATION));
+
+		if (CreateProcessA(NULL, (LPSTR)cmd.c_str(), NULL, NULL, FALSE, CREATE_NO_WINDOW, NULL, NULL, &startupInfo, &processInfo)) {
+		WaitForSingleObject(processInfo.hProcess, INFINITE);
```
bostick, line 1281: I would indent the inside of if statement

osdep/WindowsEthernetTap.cpp
```diff
@@ -1250,6 +1256,88 @@ void WindowsEthernetTap::_setRegistryIPv4Value(const char *regKey,const std::vec
...
+		CloseHandle(processInfo.hThread);
+		}
+}
+
+void WindowsEthernetTap::_allowICMPv4_cleanup(std::string addr)
+{
+		// removes 1 icmp firewall rule
+
+		char nwString[32] = { 0 };
+		sprintf(nwString, "%.16llx", _nwid);
```
bostick, line 1293: The format string is probably fine to leave as-is because this is very platform-specific code, but I try to use the ugly PRIx64 macro from <cinttypes> to be completely general, e.g., if this were to ever exist on a platform where long long is not 64-bits:
```
#include <cinttypes>
...
sprintf(nwString, "%.16" PRIx64, _nwid);
```

laduke: Thanks. I have only a foggy idea of how strings work and that section specifically was annoying me.

laduke:
but in general how do would you feel about zerotier poking your firewall thusly

glimberg:
@laduke I'm OK with it poking the firewall.  The first thing people try when setting up a network is ping and on Windows that always fails due to defaults.

laduke:
I redid it and it works better now.

PR #1842, opened 2023-01-11
cargo update tokio >= 1.24

someara:


glimberg:
Taking a look.  Pretty sure tokio is included from a 3rd party lib, so may need to update some of those first

PR #1848, opened 2023-01-12
Client & Central Controller updates to support additonal OIDC features

glimberg:
Client side:
* Fix compatibility with OneLogin
* Requested scopes vary by OIDC provider.  Different providers have different 

Controller side:
* Update Postgres queries to latest Central schema
* Added Central Controller support for the different providers
* Base OIDC provider details are still attached to an org.  Client ID & group/email lists are now associated with individual networks.


glimberg:
@adamierymenko If you could give the Network Config changes under `node/` a review it'd be appreciated.  All in working state right now but a once over wouldn't hurt

node/NetworkConfig.cpp
```diff
@@ -424,6 +425,11 @@ bool NetworkConfig::fromDictionary(const Dictionary<ZT_NETWORKCONFIG_DICT_CAPACI
 					if (d.get(ZT_NETWORKCONFIG_DICT_KEY_CLIENT_ID, this->ssoClientID, (unsigned int)sizeof(this->ssoClientID)) > 0) {
 						this->ssoClientID[sizeof(this->ssoClientID) - 1] = 0;
 					}
+					if (d.get(ZT_NETWORKCONFIG_DICT_KEY_SSO_PROVIDER, this->ssoProvider, (unsigned int)(sizeof(this->ssoProvider))) > 0) {
+						this->ssoProvider[sizeof(this->ssoProvider) - 1] = 0;
+					} else {
+						strncpy(this->ssoProvider, "default", sizeof(this->ssoProvider));
```
adamierymenko, line 431: So `strncpy` doesn't actually guarantee that the char[] ends in a null since it will fill the entire array with characters and then stop. Yet another way that C string handling is about as safe as untethered free climbing while drunk. Need to `this->ssoProvider[sizeof(this->ssoProvider) - 1] = 0` or something like that.

adamierymenko: Actually this particular line might be fine because "default" always fits.

glimberg: Went ahead & added the null anyway

PR #1851, opened 2023-01-13
Fix Possible misuse of comma operator here

bostick:
Xcode warns about "Possible misuse of comma operator here". Comma is a sequencing operator in C++ and original code does work, but is highly non-idiomatic.


someara (approved):
This is much clearer. Thanks!

Issue #1858, opened 2023-01-24
Connection state notification support?


oskarirauta:
HI.

I was wondering if it would be possible to have a notification about connection state changes, maybe as simple as notify if connection is either up or down?

I thought about changing sources by myself, but code seems a bit complex, this would ofcourse suite mostly linux (console) users but I thought about adding support for command-line argument about shell script location, and when ever service goes up or down, or program exits- it would execute the script with simple up or down argument to script, like this:
`/usr/bin/zerotier-change up`

But I couldn't find a clear place on the code where to determine that connection is succesfully up- or dropped..

This was just a simple method to do so, propably there would be better solutions- but please, avoid anything that needs huge depencies such as dbus- but something like this would be nice, since hotplug system on linux does not register when zerotier interface is up or down. Or could hotplug system be triggered directly as a notification somehow maybe?

Could something like this be though about?

laduke:
Thanks for writing. Can you explain a little about what you want to happen when an interface comes up? just for more context.

oskarirauta:
Sure thing, actually I even already wrote something like that, but haven't had a possibility to test it (and even cannot do that in few next days propably as I am just leaving for a while..)

And as a reminder, this is just something that I wrote really quick, so nowhere perfect.. But check it out for a sample what I am trying to say..

https://github.com/oskarirauta/ZeroTierOne/commit/00c369d7cb627133453dd0884aa737daa27eada5

So, can't say if it works at all, or since I am not very familiar to Zerotier project's internal code, not sure if it triggers correctly until I try- but would rather like to see someone do it the *right way* :)

There's new parameter for cli, -n which should tell location of script to execute on connection state changes.
Concept simply is this, if you for example use -n/usr/bin/zt-notifier.sh
 - when connection is established, script /usr/bin/zt-notifier.sh is executed with argument "up"
 - when connection is closed, lost, abnormal service exit, etc (connection dies), same script is executed once again with argument "down"

Script contents then is up to user, why I am interested in this, is that I am Openwrt user and recently initial zerotier protocol support was released, yes - zerotier's been there for ages, but protocol support means that system is capable to return information about state of connection (like ip addr, and if it's up or down and so on), unfortunately this is just *initial* support, as interface seems always to be down/pending state, because hotplug.d scripts or nothing is able to tell the system that hey, we did it! We are up and running!

[...]

PR #1862, opened 2023-01-27
add network event scripting support

oskarirauta:
Based on PR #1669 of @joseph-henry and issue (made by me ) #1858

If script is used, it executes it with arguments "$state" "$networkid" "$homepath"
where state is either network_up, network_update or network_down

Script hook for network events can be enabled in either as a command-line argument:
`zerotier-one -s/path/to/script.sh` 
or from local.conf, here's a example:

```
{
	"settings":
	{
		"eventHookScriptPath": "/var/lib/zerotier-one/hook.sh"
	}
}

```
interface name and other information can be retrieved then with
zerotier-cli with -D$homepath -j listnetworks for various tasks on event.

Co-authored-by: Joseph Henry @joseph-henry 
Signed-off-by: Oskari Rauta <oskari.rauta@gmail.com>

### DIFF
--- a/one.cpp
+++ b/one.cpp
@@ -1996,6 +1996,7 @@ static void printHelp(const char *cn,FILE *out)
 
 #ifdef __UNIX_LIKE__
 	fprintf(out,"  -d                - Fork and run as daemon (Unix-ish OSes)" ZT_EOL_S);
+	fprintf(out,"  -s<path>          - path to script to execute on network events" ZT_EOL_S);
 #endif // __UNIX_LIKE__
 
 #ifdef __WINDOWS__
@@ -2012,13 +2013,13 @@ static void printHelp(const char *cn,FILE *out)
 class _OneServiceRunner
 {
 public:
-	_OneServiceRunner(const char *pn,const std::string &hd,unsigned int p) : progname(pn),returnValue(0),port(p),homeDir(hd) {}
+	_OneServiceRunner(const char *pn,const std::string &hd,unsigned int p, std::string &sp) : progname(pn),returnValue(0),port(p),homeDir(hd),scriptPath(sp) {}
 	void threadMain()
 		throw()
 	{
 		try {
 			for(;;) {
-				zt1Service = OneService::newInstance(homeDir.c_str(),port);
+				zt1Service = OneService::newInstance(homeDir.c_str(),port,scriptPath.c_str());
 				switch(zt1Service->run()) {
 					case OneService::ONE_STILL_RUNNING: // shouldn't happen, run() won't return until done
 					case OneService::ONE_NORMAL_TERMINATION:
@@ -2053,6 +2054,7 @@ public:
 	unsigned int returnValue;
 	unsigned int port;
 	const std::string &homeDir;
+	const std::string &scriptPath;
 };
 
 #ifdef __WINDOWS__
@@ -2082,6 +2084,7 @@ int main(int argc,char **argv)
 	signal(SIGTERM,&_sighandlerQuit);
 	signal(SIGQUIT,&_sighandlerQuit);
 	signal(SIGINT,&_sighandlerQuit);
+	signal(SIGCHLD, SIG_IGN);
 
 	/* Ensure that there are no inherited file descriptors open from a previous
 	 * incarnation. This is a hack to ensure that GitHub issue #61 or variants
@@ -2115,7 +2118,7 @@ int main(int argc,char **argv)
 	if ((strstr(argv[0],"zerotier-cli"))||(strstr(argv[0],"ZEROTIER-CLI")))
 		return cli(argc,argv);
 
-	std::string homeDir;
+	std::string homeDir, scriptPath;
 	unsigned int port = ZT_DEFAULT_PORT;
 	bool skipRootCheck = false;
 
@@ -2134,6 +2137,11 @@ int main(int argc,char **argv)
 #ifdef __UNIX_LIKE__
 				case 'd': // Run in background as daemon
 					runAsDaemon = true;
+					break;
+				case 's': // script to run on network events
+					if (argv[i][2]) {
+						scriptPath = argv[i] + 2;
+					} else printHelp(argv[0],stdout);
 					break;
 #endif // __UNIX_LIKE__
 
@@ -2306,7 +2314,7 @@ int main(int argc,char **argv)
 	}
 #endif // __UNIX_LIKE__
 
-	_OneServiceRunner thr(argv[0],homeDir,port);
+	_OneServiceRunner thr(argv[0],homeDir,port,scriptPath);
 	thr.threadMain();
 	//Thread::join(Thread::start(&thr));
 
--- a/osdep/OSUtils.cpp
+++ b/osdep/OSUtils.cpp
@@ -11,34 +11,34 @@
  */
 /****/
 
-#include <stdio.h>
-#include <string.h>
-#include <stdlib.h>
-#include <stdarg.h>
-#include <sys/stat.h>
-#include <stdlib.h>
-
 #include "../node/Constants.hpp"
 #include "../node/Utils.hpp"
 
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+
 #ifdef __UNIX_LIKE__
-#include <unistd.h>
+#include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
-#include <sys/types.h>
+#include <netdb.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <sys/uio.h>
-#include <dirent.h>
-#include <netdb.h>
+#include <sys/wait.h>
+#include <unistd.h>
 #endif
 
 #ifdef __WINDOWS__
-#include <windows.h>
-#include <wincrypt.h>
-#include <shlobj.h>
-#include <netioapi.h>
 #include <iphlpapi.h>
+#include <netioapi.h>
+#include <shlobj.h>
+#include <wincrypt.h>
+#include <windows.h>
 #endif
 
 #include "OSUtils.hpp"
@@ -549,6 +549,53 @@ std::string OSUtils::jsonBinFromHex(const nlohmann::json &jv)
 		}
 	}
 	return std::string();
+}
+
+void OSUtils::_hookCmd(const char* scriptPath, const uint64_t nwid, const char *homePath, ZT_VirtualNetworkConfigOperation op)
+{
+	if (! scriptPath || ! strlen(scriptPath) || nwid == 0) {
+#ifdef ZT_TRACE
+		fprintf(stderr, "no script path specified\n");
+#endif
+		return;
+	}
+	long p = (long)fork();
+
+	if (p > 0) {
+#ifdef ZT_TRACE
+		fprintf(stderr, "Running network event hook script (%s)\n", scriptPath);
+#endif
+		//int exitcode = -1;
+		//::waitpid(p, &exitcode, 0);
+	}
+	else if (p == 0) {
+		::close(STDOUT_FILENO);
+		::close(STDERR_FILENO);
+		char cmdBuf[128] = { 0 };
+		OSUtils::ztsnprintf(cmdBuf, sizeof(cmdBuf), "%s", scriptPath);
+		char nwidStr[17] = { 0 };
+		OSUtils::ztsnprintf(nwidStr, sizeof(nwidStr), "%.16llx", nwid);
+		std::string cmd;
+		switch (op) {
+			case ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP:
+				cmd = std::string(cmdBuf) + " network_up " + std::string(nwidStr) + " " + std::string(homePath) + "";
+				//::execl("/bin/sh", "sh", cmdBuf, "network_up", nwidStr, homePath, "&", (const char*)0);
+				break;
+			case ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_CONFIG_UPDATE:
+				cmd = std::string(cmdBuf) + " network_update " + std::string(nwidStr) + " " + std::string(homePath) + "";
+				//::execl("/bin/sh", "sh", cmdBuf, "network_update", nwidStr, homePath, "&", (const char*)0);
+				break;
+			case ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_DOWN:	 // Same as _DESTROY
+				cmd = std::string(cmdBuf) + " network_down " + std::string(nwidStr) + " " + std::string(homePath) + "";
+				//::execl("/bin/sh", "sh", cmdBuf, "network_down", nwidStr, homePath, "&", (const char*)0);
+				break;
+			default:
+				break;
+		}
+		// Using system instead of execl, because execl blocks connections through cli
+		if (!cmd.empty()) ::system(cmd.c_str());
+		::_exit(-1);
+	}
 }
 
 #endif // OMIT_JSON_SUPPORT
--- a/osdep/OSUtils.hpp
+++ b/osdep/OSUtils.hpp
@@ -269,6 +269,16 @@ public:
 	 */
 	static std::string platformDefaultHomePath();
 
+	/**
+	 * Executes a system script specified by user
+	 *
+	 * @param scriptPath The full path to the script that should be run on each network event
+	 * @param nwid The Network ID of the network in question
+	 * @param homePath full patch to homepath containing connections settings and data
+	 * @param op The operation performed up on the network
+	 */
+	static void _hookCmd(const char *scriptPath, const uint64_t nwid, const char *homePath, ZT_VirtualNetworkConfigOperation op);
+
 #ifndef OMIT_JSON_SUPPORT
 	static nlohmann::json jsonParse(const std::string &buf);
 	static std::string jsonDump(const nlohmann::json &j,int indentation = 1);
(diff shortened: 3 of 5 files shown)


glimberg:
I'm curious what the purpose of this is when the up/down state of networks as well as their network configurations can easily be gleaned by using the JSON output of the service itself. 

For the events called out in this change:

* `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP` - This is always fired when a network is joined, or as ZeroTier rejoins known networks on startup.  It's not fired any other time.  It's not an indication of being authorized on a network and able to communicate over it.
* `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_DOWN` - This is always fired when the client is shut down, but the client hasn't left the network.  It's not an indication of being deauthorized on a network or not being able to communicate over it for other reasons.  
* `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_CONFIG_UPDATE` - This is fired often compared to the other two, and 99.9% of the time it's just the network controller issuing a certificate update, not an actual change to the network configuration.

So the first two can be determined automatically by checking whether or not the service is running.  If it's running, all networks have received `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP` once.  It'll be received If the client is not running, then all networks have received `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_DOWN` before shutting down.  Both are useful for the internal state machine inside ZeroTier for tracking the state of the virtual network adapter assigned to the network. I guess I'm fuzzy on why those two would be useful as a hook outside of ZeroTier.

As for `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_CONFIG_UPDATE`, all the info you need can be gleaned from `zerotier-cli listnetworks -j` and examining the output:

* `status`: one of OK, ACCESS_DENIED, REQUESTING_CONFIGURATION, AUTHENTICATION_REQUIRED, NOT_FOUND, PORT_ERROR, CLIENT_TOO_OLD
* `portError`: any non-0 value indicates an error with the virtual network device


Basically, the network config JSON could be polled & monitored for changes.

So given how these all work, I'm curious about why this change is necessary when the relevant information can already be gleaned elsewhere.

Additionally, if this change is merged, it'd have to work across all platforms.  macOS, Windows, *BSD and Linux.

joseph-henry:
I'd have to agree with @glimberg . The cost/benefit ratio of this modification is too high. Is there a reason the JSON output is inadequate?

joseph-henry:
To expand further, I think the original idea of ZT calling a script is sort of a bad idea (which is why the PR was closed originally), it adds a bunch of OS specific code when an OS-native script of the language of your choice would be more portable and lower the complexity surface area of ZT.

If you can make the case that some `key:value` pair needs to be exposed in the JSON output we'd be more likely to merge that.

glimberg:
> I think the original idea of ZT calling a script is sort of a bad idea 

Yeah.  ZeroTier runs as root.  In some cases it's privileges are dropped but still retains full access to do any and all network configuration tasks in the OS.  Running a script with those privileges could be extraordinarily dangerous if the wrong person is able to edit the script.

glimberg:
Converted this PR to a draft for now so it doesn't accidentally get merged

oskarirauta:
> I'm curious what the purpose of this is when the up/down state of networks as well as their network configurations can easily be gleaned by using the JSON output of the service itself.

Sure. But it doesn't notify **when** to retrieve this information. When zerotier ifd comes up, it doesn't notify hotplug.d system. So this more or less allows notification to happen in the _other way_..

Without system like this, I was not able to integrate zerotier to network manager, for example; if connection failed on system start, networking is completely unaware that it should try again. This when it doesn't start, happens most often on environments that use mobile broadband connections or similar connections which don't come with rj45 - instead require initiation which takes time, even though it's just some seconds. If zerotier's up before wan - it is just a zombie process. I could work all this out with for example, well designed cronjobs - but that sounds quite a tinkering when compared to alternative, where I just add interface to "common" networking options and it just works, renews on wan restarts, etc...

These scripts still use json output of the service itself, but this system is there to notify system, when to attempt to retrieve this information- to tell when the information is available or has changed and further actions are necessary. It also tells how to retrieve the information (homepath) - as it would be quite difficult to otherwise match event in case where there runs multiple instances of zerotier.

> 
> For the events called out in this change:
> 
> * `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP` - This is always fired when a network is joined, or as ZeroTier rejoins known networks on startup.  It's not fired any other time.  It's not an indication of being authorized on a network and able to communicate over it.

Yes, this fires up when interface comes up. Connection is not yet ready but interface is available.

> * `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_DOWN` - This is always fired when the client is shut down, but the client hasn't left the network.  It's not an indication of being deauthorized on a network or not being able to communicate over it for other reasons.

No, but it's a good notification to tell the system that it would be better to atleast check state of connection- is it really still alive, kicking and working - this also fires, when system's network manager exits zerotier - and can be used as a verification.

> * `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_CONFIG_UPDATE` - This is fired often compared to the other two, and 99.9% of the time it's just the network controller issuing a certificate update, not an actual change to the network configuration.

In this state, details about connection are available, connection is up and running - in this state, information retrieval is supposed to happen and connection state can move from penging to online.

> So the first two can be determined automatically by checking whether or not the service is running. If it's running, all networks have received `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_UP` once. If the client is not running, then all networks have received `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_DOWN` before shutting down. Both are useful for the internal state machine inside ZeroTier for tracking the state of the virtual network adapter assigned to the network. I guess I'm fuzzy on why those two would be useful as a hook outside of ZeroTier.

Yes, I am aware of numerous ways to check these, you can even work out without cli at all using ip/ifconfig and verify with ping- but again, system get's no information of any kind - **when** to do this. What could be changed here, is that we remove all the information, meaning network id and homepath and then use different methods to retrieve this information as the most crucial thing is the notification. But while we are at it... How about easing the process a bit and giving all the necessary information for it's retrieval..? It's not like that it would be a must to use, one can write a script that.. Well doesn't do anything with arguments at all, or do anything at all - even though it's pointless... As it's optional to use those arguments to anything, system's idea is to able things. What you use it then, is up to you.

> As for `ZT_VIRTUAL_NETWORK_CONFIG_OPERATION_CONFIG_UPDATE`, all the info you need can be gleaned from `zerotier-cli listnetworks -j` and examining the output:

That doesn't work at all. I need the homepath as I have multiple instances, one for every network that I have joined and no fixed port. And still, I am not using this manually, this is not for those who want to tinker in a terminal, this is for those who need automation.

> So given how these all work, I'm curious about why this change is necessary when the relevant information can already be gleaned elsewhere.

I am open for any changes - and ideas as well. This is merely a proposal, feature like this is needed- but it doesn't have to be exactly like this; This is how I did it- works and is good, but I am confident that there could be different ways to archieve same or similar results, even better.

> Additionally, if this change is merged, it'd have to work across all platforms. macOS, Windows, *BSD and Linux.

Sure. But first we would need agreement that this has any point as what I read from your comment, I am sure that I haven't explained why this is useful. I do know how to retrieve the information, system doesn't- so that's why the os needs some help, and this would do it... I am all open for conversation about topic with high hopes of getting something similar even if someone else wants to write this completely again from scratch thinking my work is rubbish :) Hopefully it won't be all that bad.......

glimberg:
> I have multiple instances, one for every network that I have joined and no fixed port

ZeroTier can join multiple networks simultaneously.  There's no need to have multiple instances for multiple networks.  This alone solves your requirement for knowing where the home directory is and what the port is as it will always be static. From there you you can set something up to monitor the service API when something changes.  API documentation for the service can be found [here](https://docs.zerotier.com/service/v1).

No, it doesn't "tell" you when something happens.  You have to ask repeatedly & determine if there are changes worth reacting to. That will likely be your best path forward.  

With your configuration of an instance per network, you'd have to write something to keep track of your running instances, their home directories, and ports they're listening on.  This is pretty far outside the typical use case for ZeroTier.

At any rate, arbitrary script execution is a non-starter as a form of notification for security reasons.  A simple mistake in file permissions on the script could lead to `rm -rf --no-preserve-root /`, or worse.  I don't think that's an avenue of attack we wish to open to potential attackers and that's why the original PR you based this on was never merged into the mainline.

oskarirauta:
> ZeroTier can join multiple networks simultaneously. There's no need to have multiple instances for multiple networks. This alone solves your requirement for knowing where the home directory is and what the port is as it will always be static. From there you you can set something up to monitor the service API when something changes. API documentation for the service can be found [here](https://docs.zerotier.com/service/v1).

I know where to find that API documentation, but it is not important or useful for normal networking environment, only if you are looking for specs, that are in normal networking only trivial.

There is exactly a need for multiple instances if you want to control the interfaces, which is in network management in a crucial point. Here, I'll show you how it's configured...

```
root@gateway:/etc/config# cat network

config interface 'loopback'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'
	option device 'lo'

config device
	option name 'br-lan'
	option type 'bridge'
	list ports 'eth0'
	list ports 'eth1'
	list ports 'eth2'
	list ports 'eth3'
	list ports 'eth4'
	list ports 'eth5'
	list ports 'eth6'
	list ports 'eth7'
	list ports 'dummy0'

config interface 'lan'
	option device 'br-lan'
	option proto 'static'
	option ipaddr '10.0.0.1'
	option netmask '255.255.0.0'
	option ip6assign '60'

config interface 'wan'
	option proto 'qmi'
	option device '/dev/cdc-wdm0'
	option modes 'all'
	option auth 'none'
	option pdptype 'ipv4v6'
	option dhcpv4 '0'
	option dhcpv6 '0'
	option mtu '1464'
	option delay '5'
	option llp 'raw-ip'
	option reqaddress 'try'
	option reqprefix 'auto'

config interface 'podman0'
	option proto 'podman'
	option device 'cni-podman0'
	option ipaddr '10.129.0.1'
	option netmask '255.255.255.0'
	option autostart 'false'

config interface 'zt0'
	option proto 'zerotier'
	#option port '9993'
	option network_id 'network1'
	option secret 'my_secret'

config interface 'zt1'
	option proto 'zerotier'
	option network_id 'network2'
	option secret 'my_secret'

```

There is **no way** to manage it with single instance with procd (consider it as lightweight systemd or something..) - as if I use only one instance - both connections are killed on ifdown... Also there comes issues when defining interface names etc, when file is getting changed by secondary instance - so they **must** run at separate instances, for so many reasons, including firewalling. Think of these entries like separate init scripts running in their own jailed environments and let me know how you would do it without breaking the concept of configuration? I could change network_id to be a list instead of variable containing single value, but then I cannot control interface names, and they are crucial for firewalling - I have a zone that holds all my vpn, in this example there are zt0 and zt1 interfaces - but you could have any other commonly supported such as wireguard, openssl, strongswan - which all can work out with this. In my opinion, zerotier is absolutely the best solution for vpn network in so many ways, but it must hold it's setup for networking in a different location. As you see, I have even cni networking in the config..

> No, it doesn't "tell" you when something happens. You have to ask repeatedly & determine if there are changes worth reacting to. That will likely be your best path forward.

Point is to provide a out-of-the-box solution, which follows operating system's standards. Like I previously told, I could set up a cronjob to monitor base path (/tmp/lib) for any zerotier_* and _try_ to iterate them, but that... Sounds a lot of work. Not for me. But for someone who is attempting put up a system quickly and painlessly without trying to find what kind of cronjobs other people have used-

I think asking repeatedly is waste of resources when there is capability to inform of changes and what you are saying, is that every other system except debian and ubuntu should build their complete eco systems differentially to have native support for zerotier? Not that it changes anything, but [this](https://github.com/oskarirauta/packages/blob/zerotier-event-scripting/net/zerotier/files/lib/netifd/zerotier.script) is script that I built. You see there is no down section defined at all, because that is handled internally by netifd, so it missing, is not a mistake.

when in the other hand.. You could just add this to /etc/config/network:
```
config interface 'zt1'
	option proto 'zerotier'
	option network_id 'network2'
	option secret 'my_secret'

```
> 
> With your configuration of an instance per network, you'd have to write something to keep track of your running instances, their home directories, and ports they're listening on. This is pretty far outside the typical use case for ZeroTier.

Can't do that for ports, if port is not defined in configuration as -p0 randomizes port - unless I do what I previously told, set up a cronjob every 20 seconds to scan for paths and reading ports from there- but nevertheless, ports are not important as you can use homepaths instead.

> At any rate, arbitrary script execution is a non-starter as a form of notification for security reasons. A simple mistake in file permissions on the script could lead to `rm -rf --no-preserve-root /`, or worse. I don't think that's an avenue of attack we wish to open to potential attackers and that's why the original PR you based this on was never merged into the mainline.

Okay- and nearly all other vpn providers think different? I see what you mean here, someone who is not talented enough, does not secure his system properly and it leads to what you described. Yet, you want to make use pesky scripting possibilities to elsewhere which just run when it's unnecessary to run them (as they just run periodically when nothing triggers them); and then you are back in square one - `rm -rf --no-preserve-root /` - so either way it could happen, argument -s is completely optional, do **not** use it if you don't need it. So so many software out there provide option --debug or --verbose - but it's completely optional to use that, just as is -s. Like you said, root executes zerotier, therefor, root should not point -s to script if it's unnecessary. If it is unnecessary, -s just should not be used at all. I just can't get around the thought, that it's most lightweight solution to provide that functionality from the source (Zerotier) which **knows** when these events are triggered instead of making a system that _guesses_ that it _might_ have changed, let's check..

When I told about other vpn's that provide such functionality, I must confess - they don't do it like this. Instead they trigger a hotplug system- kernel gets notified that interface came up, but zerotier is a exception- kernel is not triggered when zerotier comes up.

With other vpn solutions, (atleast) iface class is triggered with knowledge of interface and action, where action is either up or down. It would be easy to use this system, but impossible since zerotier does not support this.

Okay, don't get me wrong- I don't want to say this to be mean, but the impression that you are giving, is that zerotier is designed (completely) for desktop user, who hasn't seen terminal ever - there is numerous ways to mess up your system. You can use gdisk to do that, or wipefs, or any other, no you just tell that you don't want to able that, even when it's more likely that this would be abled in a different way, for example, bad root password (yes, real linuxes still have root, not just sudo, and so does freebsd..) or for example, in my case, the web interface - there's option to install terminal that can be used from web browser, and guess what - it logs in as root, as initially system holds only one user that is loggable in - root. Yet security focused system keeps doing that.

There indeed are many ways to way of destruction and I would think that abling a event script here is from the lightest end, when you think about other software, for example- samba. Or if I remember correctly even several ftpd servers provide a way to run script on logon, what if that script gets altered? And how often do you hear that it happens? Yes, proftpd should be executed as other user than root, like **every** software is suggested to- but so often it won't happen.

I think your theory behind your decision is merely reaching. But it's ok, this was my contribution to the project, glad I went through it, and without all this chattery here, I would also accepted just plain no- as I cannot seem to be able to sell you the idea and benefit of this, it goes both ways, you never can sell your idea on periodically asking for changes, that just doesn't sound very wise to mee, first you query for json, then change values that might have altered, but are not crucial for the network management, and then you make a sum of it with your favourite hasher, check if you have previous sum available and compare to that, and then start figuring out what possibly has changed and is it a good idea to update that information and throttle the firewall at the same time (everytime specs change, firewall reload is triggered), when my routers run with 8 cores and have gigabytes of ram, but same software is used on a boxes that might have 8mb rom, 16mb ram, and 125mhz soc cpu - and then you should have complex scripts to figure out if changes are necessary or not..... Waste of resources, strongswan is a beast, but nothing to compare when you have to write this kind of scripts which are hundreds of lines long and yet again there's wipefs in the middle by mistake like you said.. And in the end, it is not real time. To me that sounds... More or less manual tasking. You know you can setup physical interfaces as long as they are supported by system just with ip command (except for dns resolution)? But there are systems to make that unnecessary, make it automatic. And they work.
We just have now only one way; to agree to disagree - what you offer, api and periodic scripts... Make modern systems useless, and you can't see my point, or how this works- and it's fine. Better that I tried to contribute, instead of that I would had just asked someone to do it instead of myself, even though it ain't very pretty patch.

I wanted this to mainstream for one purpose; I could provide this as a patch, but it would be unlikely that I would maintain it and then it won't work with next version of zt. Also, you seem to be very skilled coders, therefor, if this could be improved, it would most likely happen. Too bad.

joseph-henry:
There are five reasons that this and its associated ticket https://github.com/zerotier/ZeroTierOne/issues/1858 cannot be accepted, each of which is enough on its own:

 1. As noted previously, this could be a potential security issue.
 2. The complexity that this introduces creates additional time costs in terms of testing and validating.
 3. This patch needs to be extended to multiple operating systems and maintained.
 4. This goes against ZeroTier's philosophy of trying to expose as few knobs as possible.
 5. We have already offered an alternative solution (polling the JSON output)

> I would also accepted just plain no

We do appreciate your effort but must politely decline this feature.

oskarirauta:
Any by the way; there's nothing that wouldn't be portable between unixes/linuxes, even works with musl. And not that I know, any other solutions require any kind of API/json output. They usually just setup the tunnels or what ever is used. And then notify that they did. Usually by kernel though, which then triggers hotplug.d - it's kind of a standard method.

oskarirauta:
> 1. As noted previously, this could be a potential security issue.
90% of other systems must then be vulnurable on this then...

> 2. The complexity that this introduces creates additional time costs in terms of testing and validating.
Complexity? Ok then..

> 3. This patch needs to be extended to multiple operating systems and maintained.
Possibly wouldn't work on Windows, works on osx, freebsd and linux - tested.

> 4. This goes against ZeroTier's philosophy of trying to expose as few knobs as possible.
API won't expose it? That exposes a whole lot in my opinion...

> 5. We have already offered an alternative solution (polling the JSON output)
It is not a solution to anything- it cannot be integrated as a real time system, it's periodical task depending on 3rd party software and would be far more complex (determination if zerotier is set down for a purpose for e.g.)

> We do appreciate your effort but must politely decline this feature.

Got it.

PR #1879, opened 2023-02-25
Allow macOS interfaces to use multiple IP addresses

laduke:
Long existing macos papercut.

We can't add multiple addresses to an interface on macOs unless we mess with the netmasks.
See the alias section of `man ifconfig`

> If the address is on the same subnet as the first network address for this interface, a non-conflicting netmask must be given.  Usually 0xffffffff is most appropriate.

before:
```
        inet 10.2.0.18 netmask 0xfffffe00 broadcast 10.2.1.255
        inet 10.2.0.28 netmask 0xfffffe00 broadcast 10.2.1.255
```

after:
```
        inet 10.2.0.18 netmask 0xfffffe00 broadcast 10.2.1.255
        inet 10.2.0.28 netmask 0xffffffff broadcast 10.2.0.28
```

Don't ask me why macos is like this, but before `ping` wouldn't work on all the addresses and now it does. 

Some zeronsd integration tests were relying on having multiple working addresses. So they were only working on linux. But the error messages didn't really know that, so it took a long time to debug. So that's one thing.


laduke:
not sure if those build failures are because of something in this change or not. Will look in to it later.

glimberg:
@laduke looks like the `}` on line 2333 needs to be moved up before the `#endif`

laduke:
continuous integration!

glimberg:
image

laduke:
teamwork makes the dream work

someara:
so is this good to merge?
the party parrot is very encouraging

PR #1885, opened 2023-03-02
Don't vendor packages for zeroidc

glimberg:
While vendoring rust dependencies can result in slightly shorter builds, it turns into build issues for Windows users trying to dig into & build the code base.  Windows users end up getting errors like this when they attempt to build on their own:

```
         cargo clean & cargo build --release --target=x86_64-pc-windows-msvc
     3>EXEC : error : the listed checksum of `D:\a\ZeroTierOne\ZeroTierOne\zeroidc\vendor\base64\LICENSE-APACHE` has changed: [D:\a\ZeroTierOne\ZeroTierOne\zeroidc\zeroidc.vcxproj]
         expected: a60eea817514531668d7e00765731449fe14d059d3249e0bc93b36de45f759f2
         actual:   7cfd738c53d61c79f07e348f622bf7707c9084237054d37fbe07788a75f5881c
```

As far as I can tell this is caused by git's CRLF settings, and the only way I've found to fix this is to have users wanting to build this on windows set `git config --global core.autocrl false` which turns off CRLF conversions across the board for their local machine in git.  Not an ideal solution by a long shot.  We shouldn't be forcing users to change system-wide settings just to build our software.


someara:
I've definitely lost multiple days hacking around this =)

glimberg:
> I've definitely lost multiple days hacking around this =)

Yep.  There's really no reason for it when you can pin the dependency versions

PR #1886, opened 2023-03-02
Fix #1883

### LINKED ISSUE #1883: Zerotier service crash after fail to set a friendly interface name
Hi, I got this problem in the last day and asked for help [here](https://discuss.zerotier.com/t/zerotier-service-shutting-down/12082/), but without lucky I keep looking for the reason, and after debug a while, I noticied the crash happening in this **line 114**:
image
luckly this functions references a very specifc string, this one.
image
and that lead me supect of this code:
https://github.com/zerotier/ZeroTierOne/blob/dc54e0e1c4a0566deb6261291691f17d4220e08d/osdep/WindowsEthernetTap.cpp#L842-L864
I failed to understand the IDA code very well, but by the source I think the problem is here:
https://github.com/zerotier/ZeroTierOne/blob/dc54e0e1c4a0566deb6261291691f17d4220e08d/osdep/WindowsEthernetTap.cpp#L850-L853
Seems `GetProperties` fails and keep the pointer empty, and right after that the code tries to access this pointer without checks.

glimberg:
Still unknown as to why, but the call to `nc->GetProperties()` can fail when setting a friendly name on the Windows virtual ethernet adapter. Ensure that `ncp` is not null before continuing and accessing the device GUID.

### DIFF
--- a/osdep/WindowsEthernetTap.cpp
+++ b/osdep/WindowsEthernetTap.cpp
@@ -850,12 +850,14 @@ void WindowsEthernetTap::setFriendlyName(const char *dn)
 					NETCON_PROPERTIES *ncp = nullptr;
 					nc->GetProperties(&ncp);
 
-					GUID curId = ncp->guidId;
-					if (curId == _deviceGuid) {
-						wchar_t wtext[255];
-						mbstowcs(wtext, dn, strlen(dn)+1);
-						nc->Rename(wtext);
-						found = true;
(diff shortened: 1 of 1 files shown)


glimberg:
Thanks to @marcussacana  for his help tracking this one down

marcussacana:
I builded from windows-crash-fix branch and tested in my machine, the issue is indeed fixed now.

PR #1890, opened 2023-03-07
[+] add e2k64 arch

Keylost:
Add e2k64 arch.
clang dumpmachine returns e2k64 target with new Linux Elbrus OS and Elbrus processors. 
Compiles and runs fine. Tested with Elbrus Linux 7.1 and Elbrus e2c3 CPU. clang version 9.0.1


someara:
thanks!

PR #1909, opened 2023-03-13
implemented noise_KKpsk0

mamoniot:
also fixed several minor error code bugs

### DIFF
--- a/zssp/src/main.rs
+++ b/zssp/src/main.rs
@@ -62,7 +62,7 @@ fn alice_main(
             },
             TEST_MTU,
             bob_app.identity_key.public_key_bytes(),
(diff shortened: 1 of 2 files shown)


someara:
Any chance we can get this passing status checks before merging?

It's failing on "cargo fmt", then cargo test is failing too

https://github.com/zerotier/ZeroTierOne/pull/1909

mamoniot:
How do I debug what is causing the check to fail?

glimberg:
@mamoniot `Details` button on the github workflows, then the internal Drone CI app.   Not sure if @someara ever gave you access to Drone or not, but it's now failing in the same spot as the GitHub Actions builds

PR #1910, opened 2023-03-13
fix binding temporary ipv6 addresses on macos

laduke:
The check code wasn't running.

I don't know why !defined(TARGET_OS_IOS) would exclude code on desktop macOS. I did a quick search and changed it to defined(TARGET_OS_OSX). Not 100% sure what the most correct solution there is.

You can verify the old and new versions with

`ifconfig | grep temporary`

plus

`zerotier-cli info -j` -> listeningOn


laduke:
let me double check on freebsd first...

laduke:
it was ok on freebsd before and is still good 👍

glimberg (approved):

PR #1914, opened 2023-03-15
Prevent path-learning loops

joseph-henry:
This is a continuation of the closed https://github.com/zerotier/ZeroTierOne/pull/1891

This patch is still a work in progress but it fixes the following conditions:

  - The `_paths` array was not always properly defragmented
  - The method of path learning lead to combinatorial explosions in some cases
  - Sometimes ZT could enter an endless loop of learning and re-learning the same paths
  
This code is not final and still needs to undergo some more testing before it's ready to merge.

### DIFF
--- a/node/Peer.cpp
+++ b/node/Peer.cpp
@@ -109,6 +109,13 @@ void Peer::received(
 						havePath = true;
 						break;
 					}
+					// If same address on same interface then don't learn unless existing path isn't alive (prevents learning loop)
+					if (_paths[i].p->address().ipsEqual(path->address()) && _paths[i].p->localSocket() == path->localSocket()) {
+						if (_paths[i].p->alive(now)) {
+							havePath = true;
+							break;
+						}
+					}
 				} else {
 					break;
 				}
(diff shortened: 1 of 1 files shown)


glimberg:
Converted to a draft PR so it doesn't get merged yet

erikh:
Hi, I hope my ideas are welcome.

I looked at the code, and while I am not intimately familiar with the ztone source, a cursory read suggests this is largely based on the age of the path and not the content of the path. I'm wondering, if I am correct, if coming up with a novel method of digesting the path, and keeping that on the side (e.g., some kind of data set), uninvolved in the routine operations of working with the path for any exception of computing it, would be a better solution?

To summarize because I'm a little tired, I'm thinking of keeping an active list of the paths, and a dormant hash table of the seen paths, so that in the event a path is recomputed, you can gain history on why the path was discarded previously. It might be more work in the short term, but I hope I'm communicating that in the long term, this code may need less maintenance.

HTH, and I'm sorry if I'm off-base, just kind of bored and thought I could help.

joseph-henry:
Thanks, Erik. That's something we could consider for the V2 code.

Issue #1351, opened 2021-02-15
Huge Memory Leak


gabefair:
## Required

- What is actually happening?
  -  ZeroTier has a memory leak that consumes all Swap RAM until disk is full.
- Any steps to reproduce the error.
  - Download and clean Install ZeroTierOne for Windows but do not login to any accounts.
  - Just let it run minimized in the taskbar for 3 days.
  - Do not restart or shutdown your windows computer. Just let ZeroTier run.
- Screenshot
![image](https://user-images.githubusercontent.com/2096785/107893839-6bd68680-6efb-11eb-8ea1-323d6a9e2b66.png)

## Additional information 

**Desktop (please complete the following information):**
 - OS: Windows 10 pro x64 (20H2) [Build 10.0.19042.804] 
 - ZeroTier Version: 1.6.3
  

**Additional context**
- ZeroTier Network Configuration: None. Not configured or logged in.

adamierymenko:
We'll see if we can reproduce this. In the meantime are you doing any of the following?

 * Bridging
 * Sending a lot of multicast traffic
 * Using non-IP protocols like IPX (e.g. for old games)

Thanks.

gabefair:
I had not logged into or configured zerotier. I just had installed it and forgot about it.

But on this computer I was:
1) Using Tailscale
2) Using Teamviewer
3) Using RDP
4) Using VMWare Player

justledbetter:
I've been experiencing this same issue since 1.6.0 - The service randomly spins up to 100% of CPU and consumes gigs of virtual memory; On a mix of Windows Server 2012R2, Hyper-V Server 2016, and Hyper-V Server 2019.

justledbetter:
I don't believe it happens with no-networks joined - and I think it may stop happening if IPv6 is disabled on the Windows host.  (I've only run into this on my production network and have not had the spare machines to try and experiment with it, sorry!)

riaqn:
This happending to me on linux 5.12.1. Zerotier version 1.6.5. 
Consumes 100% CPU and all memory until getting killed by OOM killer.

jonathonf:
If it's happening under Linux (and I've started seeing OOM logs on a node that has been running without other issues for a few years now) then it might be related to https://github.com/zerotier/ZeroTierOne/issues/1318 ?

danmanners:
[...]

PR #1946, opened 2023-03-29
Release memory previously allocated by UPNP_GetValidIGD

joseph-henry:
I wonder if this could be behind those unexplained memory leak reports like https://github.com/zerotier/ZeroTierOne/issues/1351


laduke:
I'll test it out

laduke:
that push is just rebasing on `dev` 
seems to work on my lan with miniupnpd.

Issue #1940, opened 2023-03-26
Behavior of exit path on 1.10.6 changed for bad


leleobhz:
Hello!

I have a VPN concentrator (I use Zerotier to access and route traffic to bunch of other VPNs like OpenVPN) and after change from 1.10.3 to 1.10.6 i noticed my connections stopped working on local network and information about exit IP on my.zerotier.com now points to Ips of any VPNs available, but not mine.

I already tried to blacklist VPN prefixes, configuring custom balancer with only my exit interface and even interface blacklist for vtuns, buty notting worked well. Apparently ZT ignores default route for some reason and try to use other interface even without route to internet over there. 

I suspect something on commit https://github.com/zerotier/ZeroTierOne/commit/ce9330e9bb06afc7047fd016d4f9882f4288b9c1 may be affecting this routing election.

What is possible to do in this case?

laduke:
Thanks for writing. A basic full tunnel mod is still working for me. 

Can you check `zerotier-cli info -j` for your listeningAddresses and surfaceAddresses? is there any difference? Does your routing table look different than from what 1.10.3 was creating? What operating system?

leleobhz:
Hello @laduke 

The only difference my `listeningAddresses` and `surfaceAddresses` have is my public IP address.

At this machine, I do not receive routes neither addresses - instead I set it manually - so there is no routing changes.

A thing I noticed is while I ping one machine to another over ZT, outside ZT interface I see traffic for both machines and in both directions, compatible com a echo request and a echo reply, but if I capture on interface, at machine noone can ping, I see both icmp getting in and out, but on other machines I dont see the awnser. 

Also happens with NDP and ARP - affected machine see both in and out traffic, but clients of this server, does not.

This server is a VyOS running zerotier with podman. Following is the podman inspect zerotier:

```[
     {
          "Id": "a2538751863af67cd72a26cf446f9ff3bef980e539fb4e036e580877c3fe01df",
          "Created": "2023-03-27T11:29:17.245650061-03:00",
          "Path": "/entrypoint.sh",
          "Args": [
               "/entrypoint.sh"
          ],
          "State": {
               "OciVersion": "1.0.2-dev",
               "Status": "running",
               "Running": true,
               "Paused": false,
               "Restarting": false,
               "OOMKilled": false,
               "Dead": false,
               "Pid": 146697,
               "ConmonPid": 146694,
               "ExitCode": 0,
               "Error": "",
[...]

PR #1961, opened 2023-04-12
Fix multiple network join from environment entrypoint.sh.release

leleobhz:
Hello!

This is a bugfix on feature "Join networks from environment" in Docker image. 

I have this patch in production and built the following way

```
podman buildx build -t quay.io/pqatsi/zerotier:latest -t quay.io/pqatsi/zerotier:1.10.6 --build-arg VERSION=1.10.6 -f Dockerfile.release .
```

pushed to https://quay.io/repository/pqatsi/zerotier?tab=tags

May this PR be approved? Also, does Zerotier interest on refactor this image? I did a experimental build at https://github.com/zerotier/ZeroTierOne/issues/1940#issuecomment-1486017714 to test a issue and I see this image is smaller and more enterprise-ready than current ones. I can do some also only improvements on debian-base if its the case too - like avoid multiple-layer hell.

Thanks!

### DIFF
--- a/entrypoint.sh.release
+++ b/entrypoint.sh.release
@@ -81,7 +81,7 @@ done
 if [ "x$ZEROTIER_JOIN_NETWORKS" != "x" ]
 then
   log_params "Joining networks from environment:" $ZEROTIER_JOIN_NETWORKS
-  for i in "$ZEROTIER_JOIN_NETWORKS"
+  for i in $ZEROTIER_JOIN_NETWORKS
   do
     log_detail_params "Configuring join:" "$i"
     touch "/var/lib/zerotier-one/networks.d/${i}.conf"


erikh:
changes LGTM. looks like someone didn't understand shell quoting.

I'm not sure what you mean by "multiple layer hell", as layers tend to help with caching and there are no artifacts. Also combining shell statements like the other dockerfile does tend to add complexity and make maintaining the image's Dockerfile much harder.

I don't personally care too much, but there's probably an argument for not changing something that works.

someara (approved):
LGTM

leleobhz:
> changes LGTM. looks like someone didn't understand shell quoting.
> 
> I'm not sure what you mean by "multiple layer hell", as layers tend to help with caching and there are no artifacts. Also combining shell statements like the other dockerfile does tend to add complexity and make maintaining the image's Dockerfile much harder.
> 
> I don't personally care too much, but there's probably an argument for not changing something that works.

Layers count on final image size. Also, I dont personally care about using RH or Debian base, but if some Zerotier customer cares about, this maybe an approach.

I'll make a PR and send to my quay.io to compare sizes. For this issue, thank you for merging.

erikh:
...

Layers count on final image size, but if you have to pull the entire image every time because you have everything mashed into a single image layer, then you're actually using more bandwidth.

I won't address this further, but it's a common misconception that less layers are better, but when you don't have to change anything but a shell script in the final layer, you're actually saving quite a bit of bandwidth on the second pull. That's how merkle trees (like git uses, too) work.

PR #1962, opened 2023-04-12
Improvements on Dockerfile.release

leleobhz:
Hello! 

As discussed in https://github.com/zerotier/ZeroTierOne/pull/1961 - i'm proposing the following improvements on Docker image of Zerotier:

* Change image to docker.io/library/debian:bullseye-slim (Slim variant with complete path to allow easier fetch for podman/oci-like container systems)
* Rebuilt instalation procedure to avoid unnecessary packages and trash like caches and packages that aren't used.
* Removed unecessary layers (Since ALL layers are pulled when user fetches the image) to decrease docker image size.

The image built at point of this commit is available on quay.io/pqatsi/zerotier:1.10.6 and was tested as following evidence:

```
leonardo@stargate-sg1:~$ sudo podman image ls
WARN[0000] Switching default driver from overlay2 to the equivalent overlay driver 
REPOSITORY               TAG         IMAGE ID      CREATED        SIZE
quay.io/pqatsi/zerotier  1.10.6      4e0ffa65aca0  2 minutes ago  107 MB
leonardo@stargate-sg1:~$ sudo podman ps
WARN[0000] Switching default driver from overlay2 to the equivalent overlay driver 
CONTAINER ID  IMAGE                           COMMAND     CREATED         STATUS             PORTS       NAMES
2d922526c544  quay.io/pqatsi/zerotier:1.10.6              17 seconds ago  Up 17 seconds ago              zerotier
leonardo@stargate-sg1:~$ show log container zerotier 
WARN[0000] Switching default driver from overlay2 to the equivalent overlay driver 
^M=> Configuring networks to join
^M=> Joining networks from command line: []
^M=> Joining networks from environment: [<NETWORK1> <NETWORK2>]
^M===> Configuring join: [<NETWORK1>]
^M===> Configuring join: [<NETWORK2>]
^M=> Starting ZeroTier
nohup: appending output to 'nohup.out'
^M===> ZeroTier hasn't started, waiting a second
^M=> Writing healthcheck for networks: []
^M=> zerotier-cli info: [200 info <CLIENT_ID> 1.10.6 OFFLINE]
^M=> Sleeping infinitely
leonardo@stargate-sg1:~$ sudo podman exec -it zerotier zerotier-cli info
WARN[0000] Switching default driver from overlay2 to the equivalent overlay driver 
200 info <CLIENT_ID> 1.10.6 ONLINE
leonardo@stargate-sg1:~$ 
```

And all networks are working well. 

It's possible to review and approve this PR?

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,23 +1,22 @@
 # vim: ft=dockerfile
 
-FROM debian:bullseye
+FROM docker.io/library/debian:bullseye-slim
 
 ARG VERSION
 
-RUN apt-get update -qq && apt-get install curl gpg -y
-RUN mkdir -p /usr/share/zerotier && \
-    curl -o /usr/share/zerotier/tmp.asc "https://download.zerotier.com/contact%40zerotier.com.gpg" && \
-    gpg --no-default-keyring --keyring /usr/share/zerotier/zerotier.gpg --import /usr/share/zerotier/tmp.asc && \
-    rm -f /usr/share/zerotier/tmp.asc && \
-    echo "deb [signed-by=/usr/share/zerotier/zerotier.gpg] http://download.zerotier.com/debian/bullseye bullseye main" > /etc/apt/sources.list.d/zerotier.list
-
-RUN apt-get update -qq && apt-get install zerotier-one=${VERSION} curl iproute2 net-tools iputils-ping openssl libssl1.1 -y
-RUN rm -rf /var/lib/zerotier-one
-
 COPY entrypoint.sh.release /entrypoint.sh
-RUN chmod 755 /entrypoint.sh
+
+RUN apt -y update \
+    && apt -y -o APT::Install-Suggests=0 -o APT::Install-Recommends=0 install ca-certificates curl gpg \
+    && curl -sSL "https://download.zerotier.com/contact%40zerotier.com.gpg" | gpg --dearmor > /etc/apt/trusted.gpg.d/zerotier.gpg \
+    && echo "deb http://download.zerotier.com/debian/bullseye bullseye main" > /etc/apt/sources.list.d/zerotier.list \
+    && apt -y update \
+    && apt -y install zerotier-one=${VERSION} \
+    && apt -y -o APT::Install-Suggests=0 -o APT::Install-Recommends=0 install iproute2 net-tools iputils-ping \
+    && apt -y purge ca-certificates curl gpg gpgconf \
+    && chmod 755 /entrypoint.sh \
+    && rm -rf /etc/apt/trusted.gpg.d/zerotier.gpg /var/lib/zerotier-one /var/lib/apt/lists/* /var/cache/apt/*
 
 HEALTHCHECK --interval=1s CMD bash /healthcheck.sh
 
-CMD []
 ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,21 +1,24 @@
 # vim: ft=dockerfile
+FROM docker.io/library/debian:bullseye-slim as gpg
+
+RUN apt -y update \
+    && apt -y -o APT::Install-Suggests=0 -o APT::Install-Recommends=0 install ca-certificates gpg gpg-agent curl \
+    && curl -sSL https://download.zerotier.com/contact%40zerotier.com.gpg -o /zerotier.asc \
+    && gpg --no-default-keyring --keyring /tmp/zerotier.gpg --import /zerotier.asc
 
 FROM docker.io/library/debian:bullseye-slim
 
 ARG VERSION
 
 COPY entrypoint.sh.release /entrypoint.sh
+COPY --from=gpg /tmp/zerotier.gpg /etc/apt/keyrings/zerotier.gpg
 
-RUN apt -y update \
-    && apt -y -o APT::Install-Suggests=0 -o APT::Install-Recommends=0 install ca-certificates curl gpg \
-    && curl -sSL "https://download.zerotier.com/contact%40zerotier.com.gpg" | gpg --dearmor > /etc/apt/trusted.gpg.d/zerotier.gpg \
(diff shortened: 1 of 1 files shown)


leleobhz:
The final image size gain is:

image

erikh:
Every time this is built, it's going to force the user to download another 107MB. This is my point. Merkle trees don't work that way. You can see when the docker client downloads an image, that it only downloads the layers it needs to update the image. This is why your debian image doesn't have to download every time. This is quite literally what makes docker downloads so fast.

FYI I wrote `docker build`.

erikh:
There are optimizations that could be done here; if curl and gpg are already present in the parent image, there is no point in updating the apt cache for another download, for example. I wrote this dockerfile and the entrypoint and I'm not on airs about it needing work.

But cramming it all into a single statement defeats the whole point of layers, and is a very myopic approach to "optimization" as at best, it optimizes the unpack of a 107MB tarball separated into 3-4 layers (executed in milliseconds on hardware already consuming electricity) instead of optimizing the download which is much more costly from an "enterprise" standpoint.

glimberg (changes requested):
Fine with the change to FROM, but the rest seems unnecessary.

rcoder:
I think the switch to the slim base image + addition of an explicit registry in the URL is a solid and uncontroversial change; thanks!

OTOH, even as someone who's been administering Debian systems for ~20 years now, I'm not sure I could catch any errors in that long chain of `x && y && z` shell commands, nor tell you exactly what might break at runtime when you purge the root CA certificates from an installed system.

I also believe some of the cache and "extra" dependency flags you're passing to apt are at least somewhat redundant with what the [debuerreotype](https://github.com/debuerreotype/debuerreotype) tooling used to generate the Debian "slim" containers is already doing.

There's almost certainly a fair bit of room left to squeeze down our image size and otherwise improve the container: we could use an even smaller base image, bring in a proper `init` replacement, identify other parts of the base system (esp. daemons and CLI tools we don't invoke) to prune, and I'd love to talk about that as part of a general effort.

But I'm inclined to agree with @erikh that container size isn't the single most critical issue for us, and I'm not entirely convinced that the new shell-chaining inside the provisioning script is a better approach.

So: consider me +1 on swapping in the (fully-qualified) slim base image, -1 on the rest (for now, anyway).

leleobhz:
Hello!

Wasn't only change to slim that reduced image size. Zerotier Docker image has too much RUN statement and increasing trash from other layers (Does not make sense to rm a file in second run when the file will remain on first run - as example). This layering issues is the reason people uses multistage builds, because other layers will be discharded when a new FROM is issued (And docker uses only last FROM layers). Layers that does not alter other layers aren't affected this effect, but Zerotier Docker does have this kind of issue.

Also about caching, the same RUN statement does what is needed and clean itself on one layer, so downloads will not bring any kind of trash with it.  https://github.com/wagoodman/dive can tell us a bit about this:

image

I agree with @rcoder . About init, I did not come with [tini](https://github.com/krallin/tini) implementation because Docker and Podman can use --init to handle it (And it works), so their inclusion may be controversial, but I agree a kind of handle of PID1 is necessary - and I can include it on PR if everyone think its a good idea.

About pipelining I agree and I'm making a new proposal. I use multistage to handle gpg to avoid poluting final layer.

About apt options, I think its safer to use apt flags directly than `--no-install-recommends` alone. In fact, zerotier deb package must handle their dependencies by their own, so even `iproute2`, `net-tools` and `iputils-ping` may be removed if not explicitly used by `/entrypoint.sh` or `/healthcheck.sh`. Also, slim images does not come with gpg and this is the reason this needs to be addressed on images that installs signed packages. And its a worthy tradeoff.

About "enterprise speed optimization", all layers of my proposal are smaller and besides debian images, my "unique layer" is smaller than all current layers - except for [`RUN mkd](https://github.com/zerotier/ZeroTierOne/blob/dev/Dockerfile.release#L9):

Zerotier Current:

image

My Proposal:

image

So this will effectivelly download less and even with less layers, my biggest layer int bigger than all other functional layers on current image. I defend this justify my effort to bring a higher optimization to Zerotier Container Image.

Finally about following point of @erikh :

> Every time this is built, it's going to force the user to download another 107MB. This is my point. Merkle trees don't work that way. You can see when the docker client downloads an image, that it only downloads the layers it needs to update the image. This is why your debian image doesn't have to download every time. This is quite literally what makes docker downloads so fast.

At this point I posted above, customer will download 80mb of already-cached debian-slim (And if debian images got updated, customer will have same issue about caching than using debian slim) and will download only a chang of 18mb - that precisely the zerotier apt package installed - without any other trace of manipulation. Wago can show this poiting to last layer and unselecting unmodified filter view.

I'm pushing my new proposal to be appreciated and I hope the arguments pointed here make sense now. As always, I'm open to suggestions and also, as Zerotier user and "unofficial advocate" I really desire to do all improvements I can do.

erikh:
you don't need tini if you change entrypoint to cmd, almost definitely, which will leave the bourne shell that is the default entrypoint of the debian image as the signal handler. the original `CMD []` statement was there to override a default CMD, and was a large part of the reason ENTRYPOINT was used at all. I could be wrong, but I doubt it.

a multi-stage build wouldn't be a bad idea, but good luck capturing all of the zerotier package contents reliably and over time without needing to do 90% of what you need to do in this image anyway in the target image.

assuming your investigation is correct, you're still talking about a size optimization that recording this thread in github's database has probably already exceeded, and the likelihood that you are going to camp on every change and maintain it is next to nil.

I'm going to unfollow this issue now as this is the equivalent of a "gold star for participation" PR and I guess when you're running a business it's important to keep up appearances but maybe stop camping on tailscale's communications like a jealous ex-girlfriend if you want some place to start. I'll stick to shucking code when I can help because I still care about the product.

leleobhz:
Hello,

As far unfortunatelly some developers cannot understand how software community works - with some offense against me - and the unabillity to discusse techinical implementations on thread, I'll close this PR and mantain my own docker implementation at https://github.com/ZenithTecnologia/zerotier-docker

PR #1964, opened 2023-04-14
Fix: warning: mutex '_aqm_m' is not held on every path through here […

bostick:
…-Wthread-safety-analysis]


joseph-henry:
That's part of a feature that wasn't fully implemented so it's likely safe for us to remove some of that stuff at some point.

bostick:
OK. What is "some of that stuff"? What should happen here?

joseph-henry:
Thanks for this as well! This code is gated as to never execute anyway (it's old QOS work) but it's probably a good idea to fix it regardless. Merging.

PR #1965, opened 2023-04-14
_bond_m guards _bond, not _paths_m

bostick:


joseph-henry:
Thanks. This does appear to be correct. Merging.

PR #1967, opened 2023-04-18
Update README.md

fengsxy:
There are two able in the text.


README.md
```diff
@@ -104,7 +104,7 @@ On CentOS check `/etc/sysconfig/iptables` for IPTables rules. For other distribu
...
+Users behind certain types of firewalls and "symmetric" NAT devices may not able to connect to external peers directly at all. ZeroTier has limited support for port prediction and will *attempt* to traverse symmetric NATs, but this doesn't always work. If P2P connectivity fails you'll be bouncing UDP packets off our relay servers resulting in slower performance. Some NAT router(s) have a configurable NAT mode, and setting this to "full cone" will eliminate this problem. If you do this you may also see a magical improvement for things like VoIP phones, Skype, BitTorrent, WebRTC, certain games, etc., since all of these use NAT traversal techniques similar to ours.
```
rcoder, line 107: As long as we're fixing this, I think the change should actually be from `...may not able able...` to `...may not _be_ able...`.

rcoder (changes requested):
I think there's one more tweak to be made, but this was a good catch. Thanks!

glimberg:
Fixed in #2000

PR #1974, opened 2023-04-25
Add short-term validation test workflow

joseph-henry:
This adds a validation workflow to confirm that ZeroTier can still perform basic connectivity functions with each commit. The workflow analyzes memory usage and code branch coverage. It also generates a small summary JSON blob with metrics we can track. This allows people to receive feedback on their commits in less than ten minutes,

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,57 @@
+on: [ push ]
+
+jobs:
+  build_ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+    - name: gitconfig
+      run: |
+        git config --global core.autocrlf input
+
+    - name: checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        target: aarch64-apple-darwin
+        override: true
+        components: rustfmt, clippy
+
+    - name: Set up cargo cache
+      uses: actions/cache@v3
+      continue-on-error: false
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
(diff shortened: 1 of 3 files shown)


joseph-henry:
@someara when you get a chance can you lend some insight as to why the cache step above fails randomly? It doesn't always fail and it just worked on my private repo a few minutes ago.
```
Error: .github/workflows/validate.yml (Line: 34, Col: 14):
Error: The template is not valid. .github/workflows/validate.yml (Line: 34, Col: 14): hashFiles('**/Cargo.lock') failed. Fail to hash files under directory '/home/runner/work/ZeroTierOne/ZeroTierOne'
```

.github/workflows/validate.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Install Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        target: aarch64-apple-darwin
```
laduke, line 20: just curious why the target is target: aarch64-apple-darwin

joseph-henry: Good question I meant to ask @someara about that as well. It was in his original build script I cannibalized

someara: That's a bug... I've made a PR to fix it here https://github.com/zerotier/ZeroTierOne/pull/1975

someara:
This looks great!

laduke:
do it

joseph-henry:
```
Bytes of memory definitely lost: 80
```

Test failed :tada:

glimberg:
image

joseph-henry:
Now we just need LLM github bot to come in an summarize the results in a surfer tone

> Bro, that was sick. You lost like 80 bytes

PR #1981, opened 2023-05-01
Support "ALL" to indicate that all domain names are resolved using th…

zch-soft:
I add a magic word "ALL" as domain of the DNS configuration to indicate that all domain names are resolved using the specified DNS server on windows.


zch-soft:
I combine this function with routing rules to make a certain node access the internet completely through another node without using the local DNS, because the local DNS may have errors and cause  failure to access Internet

glimberg:
OK.  I'm still not quite following what you're trying to do.    I need more info on what you're attempting to accomplish with this PR, as well as implementations for macOS and Linux platforms before merging.

glimberg (changes requested):
Needs macOS & Linux implementations

PR #1990, opened 2023-05-03
Brenton/thread safety fixes

bostick:
@joseph-henry Please take a look at this and let me know if anything should be changed.


joseph-henry:
Thanks. Let me get some multipath tests set up before I merge this.

node/Peer.hpp
joseph-henry, line 1: I think this makes sense.

But I also think there's a larger problem since the bond object is used in IncomingPacket without a lock. Previously this was okay since a bond was never destroyed once created, but with recent uncommitted changes this will no longer be the case. I'd accept this but it might get re-worked sometime later.

node/Peer.cpp
joseph-henry, line 1: I don't disagree with these, they would reduce performance but with upcoming changes the bond lifetime will change making these reasonable. I'd accept these changes but they might be re-worked soon anyway.

node/Bond.hpp
joseph-henry, line 1: We don't want to add a lock here since this method can be triggered by the service code. Allowing the CLI to lock a critical path (operations on `_bonds`) would not be ideal.

node/Bond.cpp
joseph-henry, line 1: I don't think this is needed because:

 - We don't want to lock critical paths (operations on `_flows`) to log a status summary
 - Reading the size is a thread safe operation

joseph-henry:
Unfortunately this is a mixed bag for me. Some of these changes I wouldn't want to include but some of them that I would include will also likely be re-worked after an upcoming patch. I'll leave it up to you as to whether you'd like to trim this down and re-submit since you did take the time to discover these issues.

bostick:
We should probably talk about what member variables are protected by what locks, what can be read / write unprotected, and what trade-offs we want.

joseph-henry:
Enough has changed that this is out of date

PR #1991, opened 2023-05-03
fix type signatures

bostick:
@joseph-henry Please take a look


joseph-henry:
These are only for reporting and sub-ms accuracy isn't needed for reporting to the UI but it is a consistency issue so we'll merge this anyway. Thanks for cleaning this up.

PR #1992, opened 2023-05-03
fix memory leak

bostick:


glimberg (approved):

joseph-henry:
I agree, this is definitely a memory leak. Good find.

PR #2005, opened 2023-05-17
Capture by-value instead of by-reference

bostick:


glimberg:
No.  That's by reference on purpose.  Some values it's referencing may change while running

PR #2011, opened 2023-05-19
remove rustc --release flag

mchangrh:
The `--release` flag should be [applied to cargo, not rustc](https://stackoverflow.com/questions/45593026/unrecognized-option-release-when-compiling-with-rustc-in-release-mode). This causes the compiler to exit with code 1 since release is an unrecognized option.

Docker mcve against version 1.10.5 with [alpine](https://gitlab.alpinelinux.org/alpine/aports/-/commit/7f7b355d081655ed3fbbc6391f6aac987d2d4e2c)
```dockerfile
FROM alpine:edge

RUN apk update && \
    apk add --no-cache alpine-sdk wget && \
    addgroup root abuild

RUN mkdir "/root/.abuild"
RUN abuild-keygen -an
WORKDIR /build
RUN wget https://gitlab.alpinelinux.org/alpine/aports/-/raw/eab757ac9e60bc25cf5d0c5b07785b12e7a4bb04/community/zerotier-one/APKBUILD && \
    wget https://gitlab.alpinelinux.org/alpine/aports/-/raw/eab757ac9e60bc25cf5d0c5b07785b12e7a4bb04/community/zerotier-one/zerotier-one.initd
RUN abuild -F checksum && abuild -F -r
```

the patch can also be tested in docker
APKBUILD
```patch
source="$pkgname-$pkgver.tar.gz::https://github.com/zerotier/ZeroTierOne/archive/refs/tags/$pkgver.tar.gz
	$pkgname.initd
+	rustc-release.patch
	"
```

rustc-release.patch
```patch
diff -Nurp a/C/make-linux.mk b/C/make-linux.mk
--- a/make-linux.mk	2023-05-19 13:12:27.946482400 -0400
+++ b/make-linux.mk	2023-05-19 13:19:00.633535900 -0400
@@ -71,7 +71,7 @@ else
 	CXXFLAGS?=-O3 -fstack-protector
 	override CXXFLAGS+=-Wall -Wno-deprecated -std=c++17 -pthread $(INCLUDES) -DNDEBUG $(DEFS)
 	LDFLAGS=-pie -Wl,-z,relro,-z,now
-	RUSTFLAGS=--release
+	RUSTFLAGS=
 endif
 
 ifeq ($(ZT_QNAP), 1)
```

### DIFF
--- a/make-linux.mk
+++ b/make-linux.mk
@@ -62,7 +62,6 @@ ifeq ($(ZT_DEBUG),1)
 	override CXXFLAGS+=-Wall -Wno-deprecated -g -O -std=c++17 -pthread $(INCLUDES) $(DEFS)
 	ZT_TRACE=1
 	RUSTFLAGS=
-	CARGOFLAGS=
 	# The following line enables optimization for the crypto code, since
 	# C25519 in particular is almost UNUSABLE in -O0 even on a 3ghz box!
 node/Salsa20.o node/SHA512.o node/C25519.o node/Poly1305.o: CXXFLAGS=-Wall -O2 -g -pthread $(INCLUDES) $(DEFS)
@@ -72,8 +71,7 @@ else
 	CXXFLAGS?=-O3 -fstack-protector
 	override CXXFLAGS+=-Wall -Wno-deprecated -std=c++17 -pthread $(INCLUDES) -DNDEBUG $(DEFS)
 	LDFLAGS=-pie -Wl,-z,relro,-z,now
-	RUSTFLAGS=
-	CARGOFLAGS=--release
+	RUSTFLAGS=--release
 endif
 
 ifeq ($(ZT_QNAP), 1)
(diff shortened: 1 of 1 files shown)


glimberg (changes requested):
This change will prevent `--release` from being passed to the actual cargo build process.

https://github.com/zerotier/ZeroTierOne/blob/da71e2524c343f563ea9c37e9668eec7c5719aaa/make-linux.mk#L414

mchangrh:
> This change will prevent `--release` from being passed to the actual cargo build process.
> 
> https://github.com/zerotier/ZeroTierOne/blob/da71e2524c343f563ea9c37e9668eec7c5719aaa/make-linux.mk#L414

I have added `CARGOFLAGS` to explicitly pass options to cargo since `RUSTFLAGS` is passed to all compilers.

glimberg (changes requested):
i'm really not sure what your goal is here.  The RUSTFLAGS variable in the makefile is only meant to pass flags to the cargo build process so that it will build a release library, or a debug library when compiling a debug version of ZeroTier

glimberg:
It's also works as is before this PR on all platforms we build for

mchangrh:
> i'm really not sure what your goal is here.

building zerotier on alpine linux fails since `--release` is passed in RUSTFLAGS to rustc, which causes it to fail with `error: Unrecognized option: 'release'`.

`--release` should only be passed to cargo or cargo-rustc, not in `RUSTFLAGS`.

glimberg:
Not sure what the difference is with Alpine vs the rest of the Linux distribution world, but this last change seems to change the intent, which is just a variable to hold flags to pass to cargo

mchangrh:
I'm currently testing between debian and alpine to try and figure out why the build is failing on one but not the other. The RUSTFLAGS environment variable/ config key is used to [pass flags to the compiler](https://doc.rust-lang.org/cargo/reference/config.html), not to cargo. The use of the RUSTFLAGS variable to pass the argument to cargo confused me.

```toml
rustflags = ["…", "…"]        # custom flags to pass to all compiler invocations
```

glimberg:
closing due to lack of activity from the reporter.

PR #2013, opened 2023-05-24
Add delay to validation test

joseph-henry:
For some reason adding the delay on line `126` seems to fix the node online check portion of the test. Thinking there still may be some race condition somewhere in the new http control plane code.


laduke:
I might poke at it a little too. This test isn't the only user who will be spamming the cli right on start up.

PR #2027, opened 2023-06-14
fix default route on mac

laduke:


laduke:
I haven't tried this on arm.

vint2k:
Hi @laduke 
Tested on m2 pro (ventura 13.4.1). It works perfect!

osdep/ManagedRoute.cpp
```diff
@@ -457,70 +471,66 @@ bool ManagedRoute::sync()
...
 
-	// Find lowest metric system route that this route should override (if any)
-	InetAddress newSystemVia;
-	char newSystemDevice[128];
-	newSystemDevice[0] = (char)0;
-	int systemMetric = 9999999;
 	std::vector<_RTE> rtes(_getRTEs(_target,false));
+
+	bool hasRoute = false;
+	rtes = (_getRTEs(_target,false));
```
rcoder, line 477: This seems redundant with line 474 above unless I'm missing something? (Which is 100% plausible.)

osdep/ManagedRoute.cpp
```diff
@@ -457,70 +471,66 @@ bool ManagedRoute::sync()
...
+				}
 			}
+			if (!newSystemDevice[0])
+				newSystemVia.zero();
+
+
+			// update the system via in case it changed out from under us
+			// while we were in default route mode
+
+			_systemVia = newSystemVia;
```
rcoder, line 519: Combined with the check on line 512 above, this looks to me like it will invoke the route editing commands with a null pointer as the value for `_systemVia` if `newSystemDevice[0]` is still false-y. The previous code appears to have simply skipped updates in the case that there was no new value for `_systemVia`, rather than calling `_routeCmd` with one or more null arguments.

Is that the desired behavior? If not, would it make sense to change the check on line 512 to skip the rest of the updates, rather than continuing with one or more empty values?

(Again, not at all sure I understand the intent perfectly, so the answer may well be "yep, working as designed!" in which case this stand as-is.)

rcoder (commented):
Asked some dumb questions based on a read-through while having relatively little idea of what's expected to happen in this area of the service code.

osdep/ManagedRoute.cpp
laduke: Thanks. That's an silly one.

rcoder:
Latest update looks good. :shipit:

PR #2052, opened 2023-07-14
Don't apply default route until zerotier is "online"

laduke:
I was running into issues with restarting the zerotier service while "full tunnel" mode is enabled.
When zerotier first boots, it gets network state from the cache on disk. So it immediately applies all the routes it knew about before it shutdown.
The network config may have change in this time.
If it has, then your default route is via a route
you are blocked from talking on. So you  can't get the current network config, so your internet does not work.
That's one theory. In any case,  starting zerotier with full tunnel enabled can fail and this helps. 

Other options include
- don't use cached network state on boot
- find a better criteria than "online"

Looking for any thoughts on how this might break other things


laduke:
debate me

glimberg:
no

Issue #2056, opened 2023-07-17
ZeroTier should exit upon loading of invalid identity.public/identity.secret


glimberg:
Current behavior is to generate a new ID when this happens.  This can be problematic in a few cases:

* Running within Kubernetes with ephemeral disks
* Running with identities mounted via read-only filesystem

In these cases, ZeroTier will continue happily running with a new identity it generates, however the identity will never be able to recovered once it exits.

PR #2058, opened 2023-07-17
Exit if loading an invalid identity from disk

glimberg:
Previously, if an invalid identity was loaded from disk, ZeroTier would
generate a new identity & chug along and generate a brand new identity
as if nothing happened.  When running in containers, this introduces the
possibility for key matter loss; especially when running in containers
where the identity files are mounted in the container read only.  In
this case, ZT will continue chugging along with a brand new identity
with no possibility of recovering the private key.

ZeroTier should exit upon loading of invalid identity.public/identity.secret #2056


glimberg:
hrmm.  That's strange @laduke.  Will look at that. 

Does work as expected with a valid-looking yet still invalid identity.secret

laduke:
replaced the identity.secret with junk and got

```
./zerotier-one: fatal error: invalid identity loaded from disk. Please remove identity.public and identity.secret from /Library/Application Support/ZeroTier/One and try again
```

laduke (approved):

rcoder:
So this feels like a small but possibly, _maybe_, for _some_ really weird usage, a breaking change. At minimum, the new exit status + messages mean that some shell scripting of the CLI could break.

If we have a good place to land a regression test for this that would be awesome. Regardless, I do want to make sure we explicitly capture it in the changelog for 1.12 due to the above.

glimberg:
> So this feels like a small but possibly, maybe, for some really weird usage, a breaking change. At minimum, the new exit status + messages mean that some shell scripting of the CLI could break.

It is an additional case where the service exits where it didn't previously.  However it now also gives *more* information to the exit error message than before when there's an exception coming from `Node`.  Previously you got "unexpected exception in main thread: unknown exception".

I'll look into getting a regression test in @joseph-henry 's  validation suite for regression testing on this

glimberg:
Ok regression test for the new behavior is now included.

rcoder:
Thanks for adding the test case! LGTM

PR #2069, opened 2023-07-25
Add search domain to macos dns configuration

laduke:
Stumbled upon this while debugging something else. If we add search domain to our system configuration for network DNS, then search domains work:

```
ping server1                                                                                                                                                                                    ~
PING server1.my.domain (10.123.3.1): 56 data bytes
64 bytes from 10.123.3.1
```

Let us know if setting search domains will cause some kind of pain or problem

### DIFF
--- a/osdep/MacDNSHelper.mm
+++ b/osdep/MacDNSHelper.mm
@@ -25,19 +25,23 @@ void MacDNSHelper::setDNS(uint64_t nwid, const char *domain, const std::vector<I
 
     CFArrayRef serverArray = CFArrayCreate(NULL, (const void**)s, servers.size(), &kCFTypeArrayCallBacks);
 
-    CFStringRef keys[2];
+    CFStringRef keys[3];
     keys[0] = CFSTR("SupplementalMatchDomains");
     keys[1] = CFSTR("ServerAddresses");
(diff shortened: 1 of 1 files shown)


osdep/MacDNSHelper.mm
```diff
@@ -25,19 +25,23 @@ static void printKeys (const void* key, const void* value, void* context) {
...
     CFArrayRef serverArray = CFArrayCreate(NULL, (const void**)s, servers.size(), &kCFTypeArrayCallBacks);
 
-    CFStringRef keys[2];
+    CFStringRef keys[3];
     keys[0] = CFSTR("SupplementalMatchDomains");
     keys[1] = CFSTR("ServerAddresses");
+    keys[2] = CFSTR("SearchDomains");
 
     CFStringRef cfdomain = CFStringCreateWithCString(NULL, domain, kCFStringEncodingUTF8);
+    CFStringRef cfdomain2 = CFStringCreateWithCString(NULL, domain, kCFStringEncodingUTF8);
```
bostick, line 34: Should there be a corresponding `CFRelease(cfdomain2)` added down below?

laduke: thanks

Issue #1696, opened 2022-06-12
MacOS DNS settings not applied


jermudgeon:
ZT 1.10.0
macOS 12.4

Behavior: setting domain and DNS servers on a network does not successfully apply to the OS
Steps to replicate:

Configure and confirm that ZT network has DNS settings enabled on client, including leave/rejoin:
`zerotier-cli listnetworks -j
<snip>
  "allowDNS": true,
  "dns": {
   "domain": "XXXXXXXX",
   "servers": [
    "XX.XX.XX.XX"
   ]
`
Confirm that domain-specific resolvers are missing or not configured with
`scutil --dns`

Note that custom resolvers can be manually added per [https://apple.stackexchange.com/questions/74639/do-etc-resolver-files-work-in-mountain-lion-for-dns-resolution/385218#385218](url)

mattbaker:
Similar position here. 

These work:
```
dig @172.22.xx.xx myhost.xz
nslookup myhost.xz 172.22.xx.xx
``

But `dig myhost.xz` and `nslookup myhost.xz` do not 🤔 

I have the DNS settings in zerotier configured with a search domain of "xz" and a server IP of 172.22.xx.xx

Very puzzling, it's like everything is their but someone zerotier is having trouble convincing macOS that it should be involved in hostname resolution

**Comment by laduke** [comment]
dig, host, nslookup,... won't work on macos for this. DNS queries go through some other system. 

try `dns-sd -G v4  myhost.xz`, `dscacheutil -q host -a name`, or use `ping`

**Comment by someara** [comment]
Hello!

This boils down to which DNS libraries a utility uses for name resolution.
Some OSX commands, like ping, will pick up the per-interface name resolution.
Others, like the bind9 utils (dig, nslookup), do not use the proper APIs, and will fail to resolve.

Try variations of these commands:

```
scutil --dns
scutil -W -r my.internal.name
```

PS: This is firmly a MacOS issue... it falls well outside the scope of ZeroTier (or ZeroNSD).

-s

mattbaker:
Interesting! The only thing that remains confusing to me is I tried to reach these hosts over ssh and via a web browser and couldn’t, I only used dig/nslookup after the fact to try to debug. 

I could reach them in the past, and trying again today I can now, so I know I had things configured correctly. What I can’t figure out is why it doesn’t work _sometimes._

I’ll use the cli tools you’ve suggested next time to debug instead of dig, maybe that will provide an interesting clue.

[...]

PR #2070, opened 2023-07-25
Fix network DNS on macOS

laduke:
It stopped working for ipv4 only networks in Monterey. See #1696

We add some config like so to System Configuration

```
scutil
show State:/Network/Service/9bee8941b5xxxxxx/IPv4
<dictionary> {
  Addresses : <array> {
    0 : 10.2.1.36
  }
  InterfaceName : feth4823
  Router : 10.2.1.36
  ServerAddress : 127.0.0.1
}

```

### DIFF
--- a/osdep/MacDNSHelper.hpp
+++ b/osdep/MacDNSHelper.hpp
@@ -12,8 +12,10 @@ class MacDNSHelper
 public:
     static void setDNS(uint64_t nwid, const char *domain, const std::vector<InetAddress> &servers);
     static void removeDNS(uint64_t nwid);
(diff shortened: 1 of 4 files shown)


joseph-henry:
Not too familiar with this but I'm curious about `addIps` and `removeIps` it only returns success if both ipv4 and ipv6 are added or removed. Is there always both to add or remove or could there just be one?

laduke:
Thanks...
That was lazy cutting a corner and I'll fix it.

laduke:
alright that's updated

Issue #1179, opened 2020-04-09
Please compile binaries without executable stack


LaurentBonnaud:
**Is your feature request related to a problem? Please describe.**

I checked if the zerotier-one binary has an executable stack and found the following:

```
$ execstack /usr/sbin/zerotier-one
X /usr/sbin/zerotier-one
```
Another way to see the problem is to look at the zerotier process:
````
# cat /proc/$(pidof zerotier-one)/maps | grep stack
7ffffa909000-7ffffa93a000 rwxp 00000000 00:00 0                          [stack]
````
**Describe the solution you'd like**

Compile the binary with a non-executable stack.  Main Linux distributions have done this work many years ago and recent toolchains do this by default.  So recompiling on a recent distribution should be enough.

**Describe alternatives you've considered**

Compiling zerotier-one oneself without executable stack is possible, but

- few people will do it
- it is more difficult to update the software
- when reporting bugs it is better to rely on the same binaries as the developer

**Comment by LaurentBonnaud** [comment]
The problem still exists in version 1.6.0 of zerotier.

**Comment by joseph-henry** [comment]
This is a pretty old ticket but I'm curious: Would `execstack -c zerotier-one` after compilation have the same effect as linking with the flag not set?

**Comment by LaurentBonnaud** [comment]
Hi,

this execstack problem still exists with the latest package version for Ubuntu 22.04/jammy:
```
Package: zerotier-one
Version: 1.8.9
```

jason-kurzik:
I encountered this today with 1.10 when trying to build for OpenSUSE Tumbleweed using their [OBS](https://build.opensuse.org/package/show/home:jason-kurzik/zerotier-one).

I found a workaround, by adding `ZT_USE_X64_ASM_ED25519=0` and `ZT_USE_X64_ASM_SALSA=0` as make parameters.  Only using 1 of them did not work for me.

I found this by comparing the output between the i586 builds and the x86_64 ones, because the i586 builds were working.

aredridel:
FWIW I've packaged 1.10.1 at https://build.opensuse.org/package/show/home:aredridel/zerotier-one by using `execstack -c` on the binary

PR #2071, opened 2023-07-25
Disable executable stacks on assembly objects

KenMacD:
Add `--noexecstack` to the assembler flags so the resulting binary will link with a non-executable stack.

Fixes zerotier/ZeroTierOne#1179


joseph-henry:
Thanks for this. I've tested it locally and it doesn't seem to break anything obvious and I'm not aware of any downsides.

Issue #2082, opened 2023-08-02
Slow startup time when zerotier starts before internet access


laduke:
edit: I can reproduce without the upnp requirement. On mac: Just turn off wifi, start zerotier, turn wifi on. 

Trying to figure out these issues https://github.com/zerotier/ZeroTierOne/issues?q=is%3Aopen+is%3Aissue+label%3ABoot%2FWake
They might not all be related. 

To reproduce something similar -this is a pain- do this:
- have a firewall that requires nat-pmp or upnp for zerotier to connect. Block incoming/returning/established UDP but not outgoing UDP. block ipv6. 
- make vm with bridged networking. i used debian 11
- log in through the vm interface, not ssh; we'll be breaking ip access to the vm. 
- stop zerotier `systemctl stop zerotier-one`
- delete the ip address from the physical interface `ip addr del 192.168.1.x/24 dev enp0s5` 
- start zerotier `/var/lib/zerotier-one/zerotier-one`
- see "sendto: network is unreachable" message from upnp library
- add ip address back to physical interface `dhclient enp0s5`
-  ping 1.1.1.1 should work
-  wait... 
- `zerotier-cli info -j` will eventually show some surface addresses, but still no direct peers in `peers`
- eventually it'll start working

this does seem worse in 1.10.6 than 1.10.1. hard to be scientific because it takes a few minutes each test. I've seen like 40 second and 4 minutes and now 1.10.6 took 15 minutes.

Some of the issues listed below say 1.10.1 or 1.8.x is better, but it's hard to tell.

`time (watch -g zerotier-cli peers | grep RELAY)`  so you don't have to stare at the terminal while it works. 

This elaborate setup is an attempt to simulate zerotier-one starting before the computer has DHCP'd its physical addresses during boot. Could this be happening on wake from sleep too? Maybe dhcp changes your physical IP when you wake from sleep and then that confuses PortMapper too. 

#2051 
#2026 
#1987 
#1958 
#1921

PR #2083, opened 2023-08-02
Fix PortMapper issue on ZeroTier startup

laduke:
See issue #2082

We use a call to libnatpmp::ininatpp to make sure the computer has working network sockets before we go into the main nat-pmp/upnp logic.

With basic exponenetial delay up to 30 seconds.

### DIFF
--- a/osdep/PortMapper.cpp
+++ b/osdep/PortMapper.cpp
@@ -14,7 +14,7 @@
 #ifdef ZT_USE_MINIUPNPC
 
 // Uncomment to dump debug messages
-//#define ZT_PORTMAPPER_TRACE 1
+#define ZT_PORTMAPPER_TRACE 1
 
 #ifdef __ANDROID__
 #include <android/log.h>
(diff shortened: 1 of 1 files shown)


laduke:
OK. This addresses a PortMapper issue, but ZeroTier is still stuck. 

I think we need to add a similar check up in OneService or Binder or something...

joseph-henry:
Updated to hopefully trigger the new workflow

joseph-henry:
```
PortMapper: started for UDP port 20001
PortMapper: NAT-PMP: request for external address failed, aborting...
PortMapper: NAT-PMP: request failed, switching to UPnP mode
Starting Control Plane...
learned new path 103.195.103.66/9993 to 778cde7190 (packet 64b14547ff432878 local socket 90113408 network 0000000000000000)
learned new path 104.194.8.134/9993 to cafe9efeb9 (packet 64d0a85186548932 local socket 90113408 network 0000000000000000)
learned new path 84.17.53.155/9993 to cafe04eba9 (packet 63613cefe73921cc local socket 90113408 network 0000000000000000)
learned new path 50.7.252.138/9993 to 62f865ae71 (packet 64724b39ab70c2a7 local socket 90113632 network 0000000000000000)
PortMapper: upnpDiscover failed, returning to NAT-PMP mode: 0
```

wongsyrone:
initnatpmp() will try to connect to `<gateway>:<NATPMP_PORT>` and fail with error code NATPMP_ERR_CONNECTERR if there is no such service on the gateway. In this case, your code treats the network as not available.

wongsyrone:
https://github.com/zerotier/ZeroTierOne/blob/27a9cc6254304c318be0f1b75fddff47210c1d2f/ext/libnatpmp/natpmp.c#L97

osdep/PortMapper.cpp
```diff
@@ -293,6 +313,7 @@ class PortMapperImpl
 #ifdef ZT_PORTMAPPER_TRACE
                     PM_TRACE("PortMapper: upnpDiscover failed, returning to NAT-PMP mode: %d" ZT_EOL_S,upnpError);
 #endif
+					break;
```
wongsyrone, line 316: This 'break' exits the `while (run)` loop

joseph-henry:
Ok, so this passes tests and it does seem to make sense. I think we're ready for a merge

Issue #1753, opened 2022-09-15
[local.conf  add   maxAcceptablePacketLossRatio ] 


daifeilail:
{
    "settings": {
        "secondaryPort": 9993,
        "tertiaryPort": 9993,
        "interfacePrefixBlacklist": [
            "docker",
            "eth"
        ],
        "defaultBondingPolicy": "balance-rr",
        "policies": {
		     
            "balance-rr": {
			    
                "upDelay": 30000,
                "downDelay": 1000,
                "failoverInterval": 500,
                "linkSelectMethod": "always",
		"maxAcceptablePacketLossRatio":0.15

            }
        }
    }
}

The above configuration can realize BOND of multiple hosts

in local.conf
Add the allowable packet loss rate For example, my use case only allows a maximum packet loss rate of 5%. more than 2 seconds. The multipath should be removed.
Until the detection is normal, the Packet will be allocated to it again
                 "upDelay": 30000,
                 "downDelay": 2000,
                 "failoverInterval": 500,

"maxAcceptablePacketLossRatio":0.05

5% packet loss rate. 2 seconds Drop line. The packet loss rate is less than 5%, and it will be redistributed to the traffic after 30 seconds

Translate with GOOGLE

Issue #1745, opened 2022-09-03
Update multipath / bonding documentation to match reality


grapexy:
From a week or two of testing ZT multipath, I've learned that following documentation will often lead to many hours of debugging, only to discover that docs were wrong.

[Multipath](https://docs.zerotier.com/zerotier/multipath/) page full of errors. Including outdated enum naming, documenting not currently implemented functionality without a proper disclaimer, documenting sample settings that are outright wrong to name a few. This should be properly updated to fix these issues.

Several issues that I've noticed:

1. Usage of built-in policy names for custom policies:
```
{
  "settings":
  {
    "defaultBondingPolicy": "balance-aware",
    "policies":
    {
      "balance-aware":
      {
        "basePolicy": "balance-aware",
      }
    }
  }
}
```
This will result in error:
`error: custom policy balance-aware will be ignored, cannot use standard policy names for custom policies`

2. Docs are using `quality` key for link quality weights, but ZT expects `qualityWeights`.
3. `balancePolicy` for `balance-aware` does not appear to be used anywhere. I.e. packet-based load-balancing is not currently implemented.
4. I may be wrong on this, but it doesn't look like link speed is taken into account for path selection at all.

In some cases, it's not clear if it's the docs that are wrong, or zt-one, so it would be nice to shine some light on the state of multipath in general. At the very least, I think it's critical to properly label or remove currently unimplemented functionality from the docs.

joseph-henry:
@grapexy Thanks again for your PR as well as general testing. There's some [new documentation](https://docs.zerotier.com/zerotier/multipath) that reflects new updates to the `dev` branch. There's still some more work in the pipeline but things should match up a little more closely now. To answer your questions:

1) This is by design. We only wanted custom behavior for non-standard policies since one might re-use a standard policy elsewhere in the config file. So, any time you tweak a parameter it'll need to be a different name.

2) This has now been fixed and is `linkQuality`. The `balance-aware` section has an [example](https://docs.zerotier.com/zerotier/multipath#balance-aware)

3) This was because `balance-aware` proved unstable in its earliest implementation and was temporarily removed. An updated implementation is now included. 

4) Correct. This was part of the `-aware` portion that was removed. It is now re-added as `capacity` and is used to hint to ZeroTier the "diameter of the pipe" in a vague sense. This will result in proportional allocation now.

[...]

PR #2086, opened 2023-08-04
Don't skip hellos when there are no paths available

laduke:
For discussion, 

working on #2082

This change helps with the issue, but i can imagine some down sides. 

This chunk of code was last changed in 

e1f60e3f8 - Behavioral changes to multipath balance modes (See: #1745 and #1753) 

right before 1.10.2 which matches some of the user's observations.

Test it out on a mac:
turn off wifi
start zerotier
turn on wifi
look at zerotier-cli peers and or try to ping a remote node

try it with and without the patch. 

We can come up with some smarter logic for deciding when to ping or not. And of course I'll remove the printfs and other junk.

### DIFF
--- a/.github/workflows/validate-1m-linux.sh
+++ b/.github/workflows/validate-1m-linux.sh
@@ -71,26 +71,6 @@ main() {
 	# Allow forwarding
 	sysctl -w net.ipv4.ip_forward=1
 
-	echo -e "\nPing from host to namespaces"
-
-	ping -c 3 192.168.0.1
-	ping -c 3 192.168.1.1
-
-	echo -e "\nPing from namespace to host"
-
-	$NS1 ping -c 3 192.168.0.1
-	$NS1 ping -c 3 192.168.0.1
-	$NS2 ping -c 3 192.168.0.2
-	$NS2 ping -c 3 192.168.0.2
-
-	echo -e "\nPing from ns1 to ns2"
-
-	$NS1 ping -c 3 192.168.0.1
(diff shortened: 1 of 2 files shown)


laduke:
Thinking about this at the moment:

https://github.com/zerotier/ZeroTierOne/blob/c6f07ee19fd78acd8a04700388684c09a1cb604c/service/OneService.cpp#L1176-L1187


later 

https://github.com/zerotier/ZeroTierOne/blob/c6f07ee19fd78acd8a04700388684c09a1cb604c/service/OneService.cpp#L1232-L1249


`_binder.refresh()` does io, but `_binder.allBoundLocalInterfaceAddresses()` and `_node->addLocalInterfaceAddress` don't, they just move some values around. 

But binder refresh is actually on a shorter timer 

`#define ZT_LOCAL_INTERFACE_CHECK_INTERVAL 60000`
`#define ZT_BINDER_REFRESH_PERIOD 30000`

that means 
`_node->processBackgroundTasks`  won't know about new physical addresses for an extra while when it tries to Hello.

joseph-henry:
Testing https://github.com/zerotier/ZeroTierOne/pull/2086/commits/704e59ce246ce5963cef46bf4083954378e8c3bd on Linux I see it reducing startup time from about `180` seconds on average to `18` seconds on average. The range being (`4` s to `25` s)

laduke:
nice. i've trimmed out all the debug stuff and force pushed.

PR #2088, opened 2023-08-09
fix macos default route again

laduke:
see commit fb6af1971 * Fix network DNS on macOS

Adding that stuff to System Config causes this extra route to be added which breaks ipv4 default route.
We found out by trial and error that adding "127.0.0.1" instead of the network'd ip address makes it work without breaking dns either. 

----
~~couldn't figure out how to fix it in SystemConfiguration so here we are~~

~~We also moved the dns helper stuff to before the syncIps stuff to help with a race condition. It didn't always work when you re-joined a network with default route enabled.~~

### DIFF
--- a/osdep/MacDNSHelper.mm
+++ b/osdep/MacDNSHelper.mm
@@ -107,7 +107,6 @@ void MacDNSHelper::removeDNS(uint64_t nwid)
 bool MacDNSHelper::addIps4(uint64_t nwid, const MAC mac, const char *dev, const std::vector<InetAddress>& addrs)
 {
     const char* ipStr = {0};
-    const char* ipStr2 = {0};
     char buf2[256] = {0};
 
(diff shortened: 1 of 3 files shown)


laduke:
I left the fprints in but commented out for reference because we want to go back and add some more optional loggin to the app and these were useful

laduke:
Here's my "test harness" 
```
watch -d -n1 "curl -m2  -s -6 ifconfig.co; curl -m2 -s -4 ifconfig.co; scutil --dns | grep lab -A3; ping -c1 drone | grep packet"
```

join 2 networks, one with a default route, one or both with ipv4 and 6 dns servers

- toggle allow default
- leave and rejoin default route network
- kill and restart zerotier 
- switch ssids (takes a few seconds to recover)
- kill and enable wifi

PR #2096, opened 2023-08-15
Add setmtu command, fix bond lifetime issue

joseph-henry:
This:

 - Adds a `setmtu` command for setting the MTU for individual bonded links

`zerotier-cli bond setmtu 1300 eth0 1.2.3.4`

 - Fixes issue where upon connectivity loss a bond can get stuck in an indeterminate state leading to one-way bonding


joseph-henry:
@bostick In this commit you'll notice some of the changes to the bond lifetime and the locks around it. This is a more stable commit to perform your analysis from.

PR #2100, opened 2023-08-21
Fix problem of doing RX while inside of TX

bostick:
Part of fix for:
https://github.com/zerotier/libzt/issues/211


joseph-henry:
Really good detective work here

PR #2107, opened 2023-08-24
Fix primary port binding issue in 1.12

joseph-henry:
The issue was that with the introduction of the new HTTP control plane we were already binding our TCP listen socket before asking the Binder to do the same. When the Binder would inevitably fail it would also take down the UDP listen sockets. I also added an IPv6 listen socket to the control plane.

I added a validator test but it's not complete, if the logic looks sound we can proceed with a merge.

### DIFF
--- a/.github/workflows/validate-linux.sh
+++ b/.github/workflows/validate-linux.sh
@@ -20,6 +20,9 @@ mkdir $TEST_DIR_PREFIX
 # How long we will wait for ZT to come online before considering it a failure
 MAX_WAIT_SECS=30
 
+ZT_PORT_NODE_1=9996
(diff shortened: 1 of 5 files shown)


joseph-henry:
On question I have is whether we should exit ZeroTier if we fail to bind on the IPv6 socket

glimberg:
> On question I have is whether we should exit ZeroTier if we fail to bind on the IPv6 socket

I'd vote no here.  It's possible that they don't have IPv6 enabled

glimberg:
Secondary control plane socket for IPv6 is now in

glimberg (approved):

PR #2110, opened 2023-08-25
Re-contact peers when default gateway changes

joseph-henry:
This combine's @laduke's default gateway detection code with a new variant of `resetWithinScope` called `resendHelloOnAllPaths` . `Node` now exposes a new function called `resetPeers()` that can be called from the service code. From my initial testing it does correctly detect changes in default gateway and doesn't crash. It likely needs some polishing but hopefully this might help a little.

### DIFF
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -1067,7 +1067,7 @@ public:
 				if (_secondaryPort) {
 					_ports[1] = _secondaryPort;
 				} else {
-					_ports[1] = _secondaryPort = _getRandomPort();
+					_ports[1] = _getRandomPort();
 				}
 			}
 #ifdef ZT_USE_MINIUPNPC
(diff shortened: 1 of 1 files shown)


laduke:
i rebased on 1.12.1 and force pushed. github seems confused. the version based on dev was borked.

joseph-henry:
I accidentally fat fingered that github review button, didn't mean to mark that as ready to review. Let me know if I need to fix anything on my end.

glimberg:
May have been [04feb30](https://github.com/zerotier/ZeroTierOne/pull/2110/commits/04feb30fbdb3c12538f9b078a09160e585502b94).  Only one with a header change that I see.  Error is:
```
          D:\a\ZeroTierOne\ZeroTierOne\osdep\PortMapper.hpp(14,10): fatal  error C1083: Cannot open include file: 'netinet/in.h': No such file or directory (compiling source file ..\..\osdep\PortMapper.cpp) [D:\a\ZeroTierOne\ZeroTierOne\windows\ZeroTierOne\ZeroTierOne.vcxproj]
         D:\a\ZeroTierOne\ZeroTierOne\osdep\PortMapper.hpp(14,10): fatal  error C1083: Cannot open include file: 'netinet/in.h': No such file or directory (compiling source file ..\..\service\OneService.cpp) [D:\a\ZeroTierOne\ZeroTierOne\windows\ZeroTierOne\ZeroTierOne.vcxproj]
```

laduke:
this is in a commit message somewhre but posting it here as the PR has changed a bunch:

After all the other changes in the service from
debugging this, this seems to not be needed.

It always reconnects for me in about 20 seconds
without resetting the peers/paths

I left the resetter code in the history if we want to try to
get it down to a couple seconds someday. It does work
but you have to poll the default gateway code
quickly and there may be other side effects.

PR #2116, opened 2023-08-28
More reliable full tunnel on macOs

laduke:
This makes switching between physical networks
with full tunnel mode enabled more reliable.
There were issues with the physical default route or device changing.


laduke:
after #2110

Issue #2121, opened 2023-09-04
Controller misses POST endpoint for updating a network configuration


vurban:
It looks like since commit e5fc89821 the controller (controller/EmbeddedNetworkController.cpp) misses the POST endpoint for updating a network configuration at /controller/network/{NWID} as documented in the [Controller API](https://docs.zerotier.com/service/v1/#operation/setControllerNetwork).
Posting valid json data to the endpoint for an existing network will always result in 404.
Sees that `EmbeddedNetworkController::configureHTTPControlPlane` is missing a call to `s.post()` for defining the endpoint.

glimberg:
This was fixed in the 1.12.1 release

vurban:
Thanks a lot for the quick response. I can confirm that updating from 1.12.0 to 1.12.1 fixed the issue.

One more remark, looks like the endpoint `/controller` for querying the status does not exist, too, but is still documented. Probably another fix in the pipe already.

PR #2122, opened 2023-09-06
fix /controller endpoint

glimberg:


rcoder:
Note: addresses this comment, which was not filed as an independent issue but did identify another needed API fix: https://github.com/zerotier/ZeroTierOne/issues/2121#issuecomment-1705511442

PR #2127, opened 2023-09-14
Fix test that was always true

bostick:
Intention was to have (x >= 0 && x <= 8) but rules[rn].v.qosBucket is unsigned (thus always >= 0)


joseph-henry:
That didn't break because of your change. Looks like an Azure/Github issue. I'll merge it once builds start to pass.

PR #2162, opened 2023-10-29
Respect user LDFLAGS

parona-source:
* Respect user LDFLAGS instead of replacing them.

Previously used on Gentoo where people want their LDFLAGS to be respected for various reasons.

https://github.com/gentoo/gentoo/blob/dbe0856cb9efde3dcd87ffb9f8b568aac2132e4a/net-misc/zerotier/files/zerotier-1.10.1-respect-ldflags.patch


joseph-henry (approved):
This seems reasonable. Thank you

Issue #2175, opened 2023-11-14
`macsrc` and `macdest` matches values are improperly parsed


lel-amri:
The rules compiler (featured in the official controller) automatically reformat MAC addresses to separate bytes with colon characters. For example, `0123456789AB` becomes `01:23:45:67:89:AB`. Matches `macsrc` (`MATCH_MAC_SOURCE` in JSON) and `macdest` (`MATCH_MAC_DEST` in JSON) are parsed in a way that doesn't ignore the `:` character, and treat it as if it was carrying address information (the current implementation treats it as a `0`). In the end, the filtering engine doesn't get the valid MAC address and thus comparisons with packet MAC addresses fails.

PR #2176, opened 2023-11-14
Fix how MAC addresses are handled by the rules parser

lel-amri:
Fixes https://github.com/zerotier/ZeroTierOne/issues/2175.

It wasn't ignoring separator characters such as the colon and hyphen. The rules compiler automatically add a colon to separate bytes, which is not compatible with how they are parsed.

---

This is currently a draft, as I'm not particularly versed in C++, and I think I couldn't fit well with the style used at Zerotier Inc.

### DIFF
--- a/controller/EmbeddedNetworkController.cpp
+++ b/controller/EmbeddedNetworkController.cpp
@@ -315,12 +315,12 @@ static bool _parseRule(json &r,ZT_VirtualNetworkRule &rule)
 		return true;
 	} else if (t == "MATCH_MAC_SOURCE") {
 		rule.t |= ZT_NETWORK_RULE_MATCH_MAC_SOURCE;
-		const std::string mac(OSUtils::jsonString(r["mac"],"0"));
+		const std::string mac(Utils::cleanMac(OSUtils::jsonString(r["mac"],"0")));
 		Utils::unhex(mac.c_str(),(unsigned int)mac.length(),rule.v.mac,6);
 		return true;
 	} else if (t == "MATCH_MAC_DEST") {
 		rule.t |= ZT_NETWORK_RULE_MATCH_MAC_DEST;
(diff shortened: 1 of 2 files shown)


laduke:
Thanks for doing this. We will try to test it soon. Do you have a quick sample ruleset that shows the problem?

lel-amri:
I think a direct and minimal test would be the following ruleset:

```
drop macsrc AA:AA:AA:AA:AA;
accept;
```

which compiles to the following JSON:

```json
{
 "rules": [
  {
   "type": "MATCH_MAC_SOURCE",
   "not": false,
   "or": false,
   "mac": "aa:aa:aa:aa:aa"
  },
  {
   "type": "ACTION_DROP"
  },
  {
   "type": "ACTION_ACCEPT"
  }
 ],
 "capabilities": [],
 "tags": []
}
```

Because of the bug, host `AA:AA:AA:AA:AA` is not filtered.

The following JSON should work though:

```json
{
 "rules": [
  {
   "type": "MATCH_MAC_SOURCE",
   "not": false,
   "or": false,
   "mac": "aaaaaaaaaa"
  },
  {
   "type": "ACTION_DROP"
  },
  {
   "type": "ACTION_ACCEPT"
  }
 ],
 "capabilities": [],
 "tags": []
}
```

joseph-henry:
Unfortunately I don't think I can merge this because the result of `cleanMac` (which is void) was being passed to the constructor of `std::string` which cannot work. I've pulled your changes from your fork and made a [new branch](https://github.com/zerotier/ZeroTierOne/pull/2244) with additional fixes. We still need to do a proper end-to-end test.

Closing this PR.

joseph-henry:
Thanks for the effort and we can still work on merging in your change with the extra fix on the other branch. Feel free to test it.

PR #2199, opened 2024-01-09
fix port rebinding issue

laduke:
It was checking for offline every loop and changing the secondary port.
But the port rebinding happens only ever X
loops. So the port would keep changing before it
would get bound.
So there was no chance to get back online.

### DIFF
--- a/service/OneService.cpp
+++ b/service/OneService.cpp
@@ -1172,25 +1172,25 @@ public:
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


joseph-henry:
Tested this on my local linux machines by blocking UDP and waiting for it to try to relay then unblocking UDP and waiting for it to rebind. In every case connection was restored in about 10 seconds. This seems to work.

PR #2210, opened 2024-01-29
Add `name` field to embedded controller networks

laduke:
Putting up here as a place to discuss. I guess we'd might want to put a limit on the length of the string? 

Storing the `name` is just this line;

`if (b.count("name")) member["name"] = OSUtils::jsonString(b["name"], "");`

this new /member2 endpoint is just a place to test, but I want to talk about that too...

We'd have to add `name` to GET /network/{nwid}/member/{ID} to as well


laduke:
This is now a 1 line diff. 

Is there any need to limit the length of `name`? The only person you'd be able to mess with with a long string as far as I can tell is your own self.

laduke:
somewhere between 4096 and 8192 bytes in the name, the HTTP server says 

`HTTP/1.1 413 Payload Too Large` 👍

PR #2211, opened 2024-01-30
Create an alternate members list endpoint

laduke:
The current api at /controller/network/1111111111767f2f/member Lists only the members' ID and revision number.
If you want details, you have to query each specific member. So if you want to make a members list, and you have 10000 members on a network, you need to make
10000 http requests.

It's also in a hard to specify and inconvenient shape:
`{ [member-id-1]: 13, [member-id-2]: 14, ... }`

This change:
GET http://localhost:9993/controller/network/1111111111767f2f/member2 ->

```
[
  {
    "authorized": false,
    "id": "1122334455",
    "name": "bob",
    ...
  },
  {
    "authorized": true,
    "id": "46254ec531",
    "name": "wakka",
    ...
  },
  {
    "authorized": false,
    "id": "2456246824",
    "name": "bob2",
    ...
  }
]
```

This currently just dumps the whole member object into the response array. We could manually specify which fields to include, but it's a bunch of fiddly c++ code to do. 

This could potentially take a few query params for sorting, filtering, and pagination. 

I'm not opinionated on `.../member2` for the endpoint name.

### DIFF
--- a/controller/EmbeddedNetworkController.cpp
+++ b/controller/EmbeddedNetworkController.cpp
@@ -872,6 +872,7 @@ void EmbeddedNetworkController::configureHTTPControlPlane(
 	std::string networkPath = "/controller/network/([0-9a-fA-F]{16})";
 	std::string oldAndBustedNetworkCreatePath = "/controller/network/([0-9a-fA-F]{10})______";
 	std::string memberListPath = "/controller/network/([0-9a-fA-F]{16})/member";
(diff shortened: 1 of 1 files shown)


laduke:
@danitheturtle leading question: thoughts on the current API at `/controller/network/{network_id}/member` ?

laduke:
Thanks for bringing up counts. That gives me an idea.

laduke:
now it's like this: 

```
{ 
  data: [ {...member1}, {...member2}, ...],
  meta: { totalCount: 4, authorizedCount: 3 }
}
```

PR #2212, opened 2024-01-31
Create an alternate network list endpoint

laduke:
Similar to recent PR for adding members2, it returns actual network objects instead of only network ids.
Also includes the total network count,
and each network has it's member counts in meta{}.

```
{
  "data": [
    {
      ...
      "id": "12345ec531867f2f",
      "objtype": "network",
      "ipAssignmentPools": [],
      "meta": {
        "authorizedMemberCount": 3,
        "totalMemberCount": 4
      },
      ...
    },
    {
    ...
    }
  ],
  "meta": {
    "networkCount": 2
  }
}
```


danitheturtle:
fantastic! Much better than it is currently. Thanks!

PR #2218, opened 2024-02-05
static file server

laduke:
this lets you host web apps out of

:9993/app/{app_name}
:9993/app/{other_app}
etc

from $ZT_HOME/app/{app_name}

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,3 +175,23 @@ Metrics are also available on disk in ZeroTier's working directory:
 | zt_peer_packet_errors | node_id | Counter | number of incoming packet errors from a peer |
 
 If there are other metrics you'd like to see tracked, ask us in an Issue or send us a Pull Request!
+
+### HTTP / App server
+
+There is a static http file server suitable for hosting Single Page Apps at http://localhost:9993/app/<app-path>
+
+Use `zerotier-cli info -j` to find your zerotier-one service's homeDir
+
+``` sh
+cd $ZT_HOME
+sudo mkdir -p app/app1
+sudo mkdir -p app/appB
+echo '<html><meta charset=utf-8><title>appA</title><body><h1>hello world A' | sudo tee app/appA/index.html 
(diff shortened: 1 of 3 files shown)


laduke:
waiting on #2220

dch:
hmm liking the look of this

joseph-henry:
Tested and this seems to work. Ready to merge?

laduke:
thanks for checking. let me remove all these per-request printfs actually.

laduke:
If you're interested in checking on a specific platform:

```sh
(start zerotier-one)
sudo mkdir -p app/hello
echo "hello" | sudo tee app/hello/index.html
curl -L localhost:9993/app/hello
```

joseph-henry (approved):
Tested and still seems to work

dch:
LGTM too here. I'm interested to know what the original & intended use case for this.

Some comments, for docs:

- the dir only needs to be readable by root, so you can `chown dch $ZT_HOME/app` and it still works just fine

- when no index.html is found, please return 404 instead of 500
   e.g.`GET /app/missing/` should return 404

- please consider documenting/allowing the the directory location configurable. 
  I tested it with a softlink and it seeems to work, but a nice config option
  would be great, you know we will be asking for it anyway ;-)

- I did basic path traversal checks, what (if any) security guarantees are on offer here?

- It seems urls with .. or %2e%2e are resolved "safely", and trying to escape with 
    http://localhost:9994/app/../../usr/share/doc/ntp/index.html or 
    http://localhost:9994/app/%2e%2e/%2e%2e/usr/share/doc/ntp/index.html all fail
    which is great

- you have some support for MIME types, e.g. `index.json` is returned as `application/json, nice

laduke:
Thanks for testing it out! 
We're relying on https://github.com/yhirose/cpp-httplib for the heavy lifting, including not allow traversal. It was already included in this codebase. 

More news and info Coming Soon

glimberg (approved):
lgtm

dch:
did this make it into https://docs.zerotier.com ? I can't find it mentioned yet.

Issue #2198, opened 2024-01-05
Service JSON API Listens on All Interfaces by default


44670:
Hello ZeroTier team,

I've encountered an inconsistency in the behavior of your service as described in the README versus its actual default behavior. The [README](https://github.com/zerotier/ZeroTierOne/blob/663ed73768a895a8681b6ec57059109073adf8f9/README.md?plain=1#L84)   states that `The service is controlled via the JSON API, which by default is available at 127.0.0.1 port 9993.` 

However, in practice, the service does not behave as described. Instead, it listens on the 9993 TCP port across all interfaces, allowing remote computers to access the service's JSON API.

```
# netstat -anp -t | grep 9993
tcp        0      0 0.0.0.0:9993            0.0.0.0:*               LISTEN      1220/zerotier-one
tcp6       0      0 :::9993                 :::*                    LISTEN      1220/zerotier-one

# curl http://192.168.1.110:9993
{}

# zerotier-cli -v
1.12.2
```

While I understand that a secret is required for actual operations, this default behavior still exposes an unnecessary attack surface and contradicts the expected behavior outlined in the README documentation. This could lead to potential security concerns for users who rely on the documented default settings.

I believe this issue warrants attention to align the actual default behavior of the JSON API with the documented expectations, thereby enhancing security and clarity for users.

Thank you for looking into this matter.

joseph-henry:
Thanks for pointing this out.

We don't completely disagree with your concerns but there are valid reasons for it to listen on `0.0.0.0:9993` with the current design (e.g. remote monitoring, controller API, etc).

I've made a PR to address the inconsistency in the docs.

PR #2219, opened 2024-02-08
Update client JSON API endpoint address in README.md

joseph-henry:
User pointed out an inconsistency in our documentation: https://github.com/zerotier/ZeroTierOne/issues/2198

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ On most distributions, macOS, and Windows, the installer will start the service 
 
 A home folder for your system will automatically be created.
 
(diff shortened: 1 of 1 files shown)


joseph-henry:
Yeah I thought about that as well since "available at" doesn't exclude other addresses but felt it was more straightforward to list what it binds to.

glimberg:
Yeah, but I think this change makes it even more unclear.

glimberg:
Now it reads like you can control the API from anywhere by default as long as you have the key.  That's not the case.

glimberg:
looks good to me

PR #2229, opened 2024-02-21
Add homeDir to info json

laduke:
Want to show it in UIs. So need to get it via API.

It's also a pain to look up. You have to go to external docs.

zerotier-cli info -j

```json
"config": {
  "settings": {
   "allowTcpFallbackRelay": true,
   "forceTcpRelay": true,
   "homeDir": "/Library/Application Support/ZeroTier/One",
```

under `settings` because it's a setting you can set when you start zerotier-one
`Usage: /var/lib/zerotier-one/zerotier-one [-switches] [home directory]`


joseph-henry (approved):
Oh yeah, I've needed this too. Thanks for adding it

PR #2233, opened 2024-02-23
ZeroTier on multiple threads

joseph-henry:
## ZeroTier on multiple threads

This alpha-quality patch enables concurrent processing of packets in the RX and TX directions and appears to improve performance significantly in many situations (but also make some worse). It automatically scales thread count according to `std::thread::hardware_concurrency()` and can optionally be overridden at start time by setting the `ZT_PACKET_PROCESSING_CONCURRENCY` environment variable. You can enable CPU pinning at start time via `ZT_CPU_PINNING=1`.

This has only been tested on Linux and FreeBSD and is not fully implemented in macOS so don't even try.

Example usage (`ZT_PACKET_PROCESSING_CONCURRENCY` is entirely optional since it'll automatically use all cores):

```
ZT_CPU_PINNING=1 ZT_PACKET_PROCESSING_CONCURRENCY=4 ./zerotier-one
```

To force ZeroTier onto one thread:

```
ZT_PACKET_PROCESSING_CONCURRENCY=1 ./zerotier-one
```


joseph-henry:
Closing due to incorrect base branch. Will re-open a new PR

PR #2234, opened 2024-02-23
Add multi-core concurrent packet processing

joseph-henry:
## ZeroTier on multiple threads

This alpha-quality patch enables concurrent processing of packets in the RX and TX directions and appears to improve performance significantly in many situations (but also make some worse). It automatically scales thread count according to `std::thread::hardware_concurrency()` and can optionally be overridden at start time by setting the `ZT_PACKET_PROCESSING_CONCURRENCY` environment variable. You can enable CPU pinning at start time via `ZT_CPU_PINNING=1`.

This has only been tested on Linux and FreeBSD and is not fully implemented in macOS so don't even try.

Example usage (`ZT_PACKET_PROCESSING_CONCURRENCY` is entirely optional since it'll automatically use all cores):

```
ZT_CPU_PINNING=1 ZT_PACKET_PROCESSING_CONCURRENCY=4 ./zerotier-one
```

To force ZeroTier onto one thread:

```
ZT_PACKET_PROCESSING_CONCURRENCY=1 ./zerotier-one
```

### DIFF
--- a/osdep/BSDEthernetTap.cpp
+++ b/osdep/BSDEthernetTap.cpp
@@ -39,7 +39,9 @@
 #include <net/if_dl.h>
 #include <net/if_media.h>
 #include <net/route.h>
+#include <pthread_np.h>
 
+#include <sched.h>
 #include <string>
 #include <map>
 #include <set>
@@ -53,6 +55,7 @@
 #include "BSDEthernetTap.hpp"
 
 #define ZT_BASE32_CHARS "0123456789abcdefghijklmnopqrstuv"
+#define ZT_TAP_BUF_SIZE (1024 * 16)
 
 // ff:ff:ff:ff:ff:ff with no ADI
 static const ZeroTier::MulticastGroup _blindWildcardMulticastGroup(ZeroTier::MAC(0xff),0);
@@ -61,6 +64,7 @@ namespace ZeroTier {
 
 BSDEthernetTap::BSDEthernetTap(
 	const char *homePath,
+	unsigned int concurrency,
 	const MAC &mac,
 	unsigned int mtu,
 	unsigned int metric,
@@ -69,6 +73,7 @@ BSDEthernetTap::BSDEthernetTap(
 	void (*handler)(void *,void *,uint64_t,const MAC &,const MAC &,unsigned int,unsigned int,const void *,unsigned int),
 	void *arg) :
 	_handler(handler),
+	_concurrency(concurrency),
 	_arg(arg),
 	_nwid(nwid),
 	_mtu(mtu),
@@ -195,11 +200,9 @@ BSDEthernetTap::BSDEthernetTap(
 BSDEthernetTap::~BSDEthernetTap()
 {
 	::write(_shutdownSignalPipe[1],"\0",1); // causes thread to exit
-	Thread::join(_thread);
 	::close(_fd);
 	::close(_shutdownSignalPipe[0]);
 	::close(_shutdownSignalPipe[1]);
-
 	long cpid = (long)vfork();
 	if (cpid == 0) {
 #ifdef ZT_TRACE
@@ -210,6 +213,10 @@ BSDEthernetTap::~BSDEthernetTap()
 	} else if (cpid > 0) {
 		int exitcode = -1;
 		::waitpid(cpid,&exitcode,0);
+	}
+	Thread::join(_thread);
+	for (std::thread &t : _rxThreads) {
+		t.join();
 	}
 }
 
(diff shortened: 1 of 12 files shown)


Sandros94:
I wonder if this could improve performance on smaller cpus like the ones in commercial NASs

joseph-henry:
*Update: Packet re-ordering seemed to be an issue in situations where a single TCP stream was being received by a large number of high-performance cores so the following changes were made which I believe are a good compromise for the time being:*

This latest commit will not have multicore enabled by default, it can be enabled with `ZT_ENABLE_MULTICORE=1`

When enabled it will only use `2` cores if at least `4` logical cores are available. No matter how many cores beyond that are present it will only use `2`. To override this you can set `ZT_CONCURRENCY=N`.

To experiment with core pinning you can use `ZT_CORE_PINNING=1` but this is most likely a bad idea so do this last.

Suggested default usage:

```
sudo ZT_ENABLE_MULTICORE=1 ./zerotier-one
```

I am interested in hearing how this performs for people.

Thanks.

joseph-henry:
> I wonder if this could improve performance on smaller cpus like the ones in commercial NASs

Yes, exactly. This is where I'm seeing the best gains in my testing.

Sandros94:
@joseph-henry sorry for the dumb question: is the [Dockerfile.ci](https://github.com/zerotier/ZeroTierOne/blob/34cc26176c8d1a2c49935928d7297a9859ab2023/Dockerfile.ci) enought to quickily test this?

Sandros94:
> I am interested in hearing how this performs for people.

I'm not entirely sure if I'm building it right but I just did some tests, in particular related to video workflows using Blackmagic Disk Speed Test.
Source was a Synology DS1522+ (zerotier built from source, no other containers nor connections active). Destination is a Win11+Ryzen 5800X (zerotier 1.14 stable) over a public network.

The connection should reach 100mbit/s from Win to Synology, and 900mbit/s from Synology to Win. All tests are from the Win machine perspective.

1. With `ZT_ENABLE_MULTICORE=0`: upload is 100mbit/s; download is ~410mbit/s
2. With `ZT_ENABLE_MULTICORE=1` and `ZT_CONCURRENCY=2`: upload is ~80mbit/s; download is ~290mbit/s
3. With `ZT_ENABLE_MULTICORE=1` and `ZT_CONCURRENCY=4`: upload is ~50mbit/s; download is ~260mbit/s

`ZT_CORE_PINNING=1` didn't make a difference, but I've also noticed during uploads that the speed is quite inconsistent
image

P.S: container running from this [Docker Hub image (tag `multicore-64634c9`)](https://registry.hub.docker.com/r/sandros94/zerotier-synology) built with [this dockerfile](https://github.com/Sandros94/zerotier-synology/blob/7731cc07777346f9c528aa414dffc4b99eca5bb2/dsm7-docker/multicore.Dockerfile).

TommyKing:
> > I am interested in hearing how this performs for people.
> 
> I'm not entirely sure if I'm building it right but I just did some tests, in particular related to video workflows using Blackmagic Disk Speed Test. Source was a Synology DS1522+ (zerotier built from source, no other containers nor connections active). Destination is a Win11+Ryzen 5800X (zerotier 1.14 stable) over a public network.
> 
> The connection should reach 100mbit/s from Win to Synology, and 900mbit/s from Synology to Win. All tests are from the Win machine perspective.
> 
>     1. With `ZT_ENABLE_MULTICORE=0`: upload is 100mbit/s; download is ~410mbit/s
> 
>     2. With `ZT_ENABLE_MULTICORE=1` and `ZT_CONCURRENCY=2`: upload is ~80mbit/s; download is ~290mbit/s
> 
>     3. With `ZT_ENABLE_MULTICORE=1` and `ZT_CONCURRENCY=4`: upload is ~50mbit/s; download is ~260mbit/s
> 
> 
> `ZT_CORE_PINNING=1` didn't make a difference, but I've also noticed during uploads that the speed is quite inconsistent image
> 
> P.S: container running from this [Docker Hub image (tag `multicore-64634c9`)](https://registry.hub.docker.com/r/sandros94/zerotier-synology) built with [this dockerfile](https://github.com/Sandros94/zerotier-synology/blob/7731cc07777346f9c528aa414dffc4b99eca5bb2/dsm7-docker/multicore.Dockerfile).

i get the same result with 50Mb/s Upload Connection
1. with ZT_ENABLE_MULTICORE=0 = Upload 5MB/s
2. with ZT_ENABLE_MULTICORE=1 and ZT_CONCURRENCY=2 = 4,4MB/s
3. With ZT_ENABLE_MULTICORE=1 and ZT_CONCURRENCY= 4 i get around 4.0 MB/s

joseph-henry:
Thanks for your results everybody. It's still a work in progress. 

Some updates:

 - Packets are sorted by flow to prevent re-ordering (though this doesn't seem to be a full solution)
 - Configuration is now done via `local.conf`, not environment variables
 
 Example config:
 
 ```
 {
    "settings":
    {
        "multicoreEnabled": true,
        "concurrency": 4,
        "cpuPinningEnabled": false
    }
}
```

More updates to come.

**Comment by joseph-henry** [comment]
Update: At this point we are seeing decent gains in most cases and no worsening of performance in any case. However, this is claim is only for Linux at the moment. Once things are fully ironed out we plan on porting these changes to other platforms.

If anyone is willing to give this another go it would be very much appreciated. Thank you

**Comment by sandros94** [comment]
Thanks @adamierymenko, just a quick question: where could I follow its development and to quickly test it in a Synology nas should I build my own docker image based on `dev` branch or is there a CI build for that?

**Comment by joseph-henry** [comment]
> should I build my own docker image based on `dev` branch or is there a CI build for that?

> Also: currently it is only configurable via configs, ENV variables are disabled still?

Yes building a docker image from latest `dev` would be best for testing this. And also yes we made the decision to configure via `local.conf` instead of ENV vars.

**Comment by sandros94** [comment]
> And also yes we made the decision to configure via `local.conf` instead of ENV vars.

Hopefully it's just a temporary decision, handling ENV vars in a docker container is such a breaze compared to binding mounts, in particular when A/B testing.

Thanks for the work

**Comment by raryanpur** [comment]
> > And also yes we made the decision to configure via `local.conf` instead of ENV vars.
> 
> 
> 
> Hopefully it's just a temporary decision, handling ENV vars in a docker container is such a breaze compared to binding mounts, in particular when A/B testing.
> 
> 
> 
> Thanks for the work

Maybe both? Having runtime configuration is super helpful.

**Comment by joseph-henry** [comment]
That's actually a good point. We'll consider doing that.

**Comment by sandros94** [comment]
For now I'm going to leave this here, since there isn't an official issue to track this, nor I do have useful information/reproducibility to open up one, but I've **quickly** tested 1.14.1 with the below `local.conf` and got half the speed compared without multicore. But still didn't experience the ups and downs I was getting last time I've reported.

Same setup as last time:
- Synology DS1522+ with DSM7 (multicore enabled)
- Win11+Ryzen 5800X

`/var/lib/zerotier-one/local.conf`:
```
{
  "settings":
  {
    "multicoreEnabled": true,
    "concurrency": 2,
    "cpuPinningEnabled": false
  }
}
```

jklop123:
> For now I'm going to leave this here, since there isn't an official issue to track this, nor I do have useful information/reproducibility to open up one, but I've **quickly** tested 1.14.1 with the below `local.conf` and got half the speed compared without multicore. But still didn't experience the ups and downs I was getting last time I've reported.
> 
> Same setup as last time:
> 
>     * Synology DS1522+ with DSM7 (multicore enabled)
> 
>     * Win11+Ryzen 5800X
> 
> 
> `/var/lib/zerotier-one/local.conf`:
> 
> ```
> {
>   "settings":
>   {
>     "multicoreEnabled": true,
>     "concurrency": 2,
>     "cpuPinningEnabled": false
>   }
> }
> ```

I tested some situations. To enjoy the performance improvement of multithreading, both ends must be Linux clients and multithreading must be enabled.

pve virtual machine, two zerotier clients, debian12, x64.
Client server does not enable multithreading: 1.5G
Client server only enables one multithreading support: 1.5G
Client server both enable multithreading: 2.9-3G, reaching the 2x performance improvement claimed by the official blog.

My confusion is that if the server opens another iperf3 instance and adds the client2 -> server test, the bandwidth of the two sets of tests is still only 3G.
I am not sure whether this result meets expectations. Maybe the bottleneck now is that the zerotier virtual network card still has only one rx tx queue? In any case, no further research was done.

ps: This test environment is a server environment, the CPU frequency is not high, so the absolute performance is not very good.

PR #2241, opened 2024-03-05
Comment out debug traces in Rules Engine

joseph-henry:
I think these are fine to leave in for now but they're too noisy for `ZT_TRACE` since it's per-packet. Maybe once we're done tweaking the rules engine we can remove them entirely


laduke (approved):
Yeah I was wondering about that. Is it a huge pain to make a new make flag? I guess commenting out is fine.

joseph-henry:
I think adding varying trace levels wouldn't be a bad idea for a future release

PR #2245, opened 2024-03-05
more project updates for win ARM64

glimberg:


laduke (approved):
this builds on my arm64 vm

rcoder (approved):
This looks good to me; AFAIK we're really the only folks to build ZT1 on Windows on the regular, so bumping the stdlib and VS versions shouldn't impact too many other folks.

glimberg:
It was just setting ARM64 Debug to match the Release build & other architectures.

PR #2252, opened 2024-03-14
Fix condition where active-backup would fail to engage

joseph-henry:
In some cases links in a bond wouldn't be marked as bonded resulting in the bonding mode not registering itself as ready. This then results in an active bonding layer that appears to be working when in fact only the classical path handling logic is running.

This can be tested with `local.conf` similar to (with correct interface names of course):

```
{
  "settings":
  {
    "allowSecondaryPort": false,
    "defaultBondingPolicy": "rapid-active-backup",
    "policies":
    {
      "rapid-active-backup":
      {
        "basePolicy": "active-backup",
        "failoverInterval": 1000,
        "links":
          {
            "enp7s0": { "failoverTo": "eth0", "mode": "primary" },
            "eth0": { "mode": "spare" }
          }
      }
    }
  }
}
```

When a cable is pulled:
 - What did happen: Sometimes the bond would report using a link when in fact it was using another and then fail to... faillover
 - What it does now: Use correct link and failover

### DIFF
--- a/node/Bond.cpp
+++ b/node/Bond.cpp
@@ -1094,6 +1094,7 @@ void Bond::curateBond(int64_t now, bool rebuildBond)
 	 * Curate the set of paths that are part of the bond proper. Select a set of paths
 	 * per logical link according to eligibility and user-specified constraints.
 	 */
+	int updatedBondedPathCount = 0;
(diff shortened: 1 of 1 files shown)


laduke (approved):
we tested across our computers and it worked

Issue #2151, opened 2023-10-18
GET http://localhost:9993/status X-ZT1-Auth:... fails over IPv4 but works over IPv6 


dch:
## Please let us know

There's a regression between 1.12.0 and 1.12.1 that breaks ipv4 localhost query
via `zerotier-cli info`. I've only just noticed while adding a new node that
this is broken.
 
Under v1.12.1:

```
# sudo zerotier-cli status
401 status {}

# zerotier-cli dump
Error connecting to the ZeroTier service: {}

Please check that the service is running and that
TCP port 9993 can be contacted via 127.0.0.1.

# sockstat -46lp 9993
USER     COMMAND    PID   FD  PROTO  LOCAL ADDRESS         FOREIGN ADDRESS
root     zerotier-o 85664 7   tcp4   *:9993                *:*
root     zerotier-o 85664 8   tcp46  *:9993                *:*
root     zerotier-o 85664 14  udp4   10.0.0.112:9993       *:*
```

But clearly the daemon *is* up, and listening.

We'd like to see, as under previous versions, the info.

Under v1.12.0:

```
> sudo zerotier-cli info
200 info 7c5005f22b 1.12.0 ONLINE
```

- environment is ansible-managed
- runs ZT since over a decade probably
- OS: FreeBSD (various versions)
- ZT: v1.12.2 (everywhere)
- `authtoken.secret` hasn't changed in years on nodes

Running the same thing using tcpdump/ngrep, we see that this is basically the same as running an HTTP `GET` to `http://localhost:9993/status` with header `X-ZT1-Auth:my_secret` so let's do that with curl:

```
$ curl -v http://localhost:9993/status -4HX-ZT1-Auth:...
< HTTP/1.1 401 Unauthorized
...
{}

$ curl -v http://localhost:9993/status -6HX-ZT1-Auth:...
< HTTP/1.1 200 OK
...
{"address":"9bbbdbfdd2","clock":1697648307359,"config":{"settings": ....}
```

The `-v4` version should be authorized correctly, but isn't, whether via `zerotier-cli` or `curl`.

laduke:
Hello! Thanks for reporting. This isn't happening on my mac; both curl -4 and curl -6 work. So that's strange. 

Are there any error messages when zerotier is starting up?

dch:
[...]

PR #2253, opened 2024-03-15
Consider ::ffff:127.0.0.1 as a loopback address

laduke:
cpp-httplib  sets IPV6_V6ONLY to false on it's sockets. On FreeBSD, this makes all ipv4 addresses get get prefixed with `::ffff:`. It makes them into IPv4 mapped IPv6 addresses.

This is a partial fix for #2151. The cli will work again. Something should probably also be adjusted with httplib.

If you want to, for example, use the `allowManagementFrom` option in local.conf
you will need to prefix it with "::ffff:", "::ffff:1.2.3.4" which is a little surprising and inconsistent between BSD and other OSs.

Thanks @dch for figuring this out. We used your example code but tried to make it match with what already exists. 

```
root@vultr:~/ZeroTierOne # uname -a
FreeBSD vultr.guest 14.0-RELEASE-p3 FreeBSD 14.0-RELEASE-p3 #0: Mon Dec 11 04:56:01 UTC 2023     root@amd64-builder.daemonology.net:/usr/obj/usr/src/amd64.amd64/sys/GENERIC amd64
root@vultr:~/ZeroTierOne # ./zerotier-cli info
200 info ad521db9aa 1.12.2 ONLINE
```

### DIFF
--- a/node/InetAddress.cpp
+++ b/node/InetAddress.cpp
@@ -132,7 +132,20 @@ InetAddress::IpScope InetAddress::ipScope() const
 					return IP_SCOPE_PRIVATE;        // fc00::/7
 				}
 			}
+
+			// :::ffff:127.0.0.1
+			// 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 0x7f, 0, 0, 1
 			unsigned int k = 0;
+			while ((!ip[k])&&(k < 9)) {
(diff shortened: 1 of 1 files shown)


joseph-henry (approved):
The parsing logic seems correct as it mirrors existing patterns in the code and I don't see any bugs, but I haven't personally tested this yet.

PR #2263, opened 2024-03-25
Update README.md minimal compilers

fedepell:
Update the minimal compilers in documentation. As `std=c++17` is used the minimal got quite bumped from last update. For gcc 7.x would seem possibly enough, but some of the dependencies (eg. libpqxx) need 8.x at least, so setting that as documented minimum.


adamierymenko (approved):
Looks reasonable.

PR #2287, opened 2024-05-09
Fix macOS GitHub Action

glimberg:
x64 rust toolchain wasn't being installed


laduke (approved):

joseph-henry (commented):
Yeah looks like we were only testing Apple silicon

glimberg:
I think GitHub may have updated their fleet of action runners to M series chips (finally)

Issue #2289, opened 2024-05-10
compatibility for android with ndk toolchain


eventlOwOp:
### problem
when cross compiling zerotier using ndk toolchain (for android)

it satisfied `#if defined(__LINUX__) && ( (!defined(__GLIBC__)) || ((__GLIBC__ >= 2) && (__GLIBC_MINOR__ >= 18)) )` on line 2097 and called `pthread_setattr_default_np` which is not supported in ndk

```
one.cpp:2097:2: error: use of undeclared identifier 'pthread_setattr_default_np'
        pthread_setattr_default_np(&tattr);
        ^
1 warning and 1 error generated.
make: *** [<builtin>: one.o] Error 1
make: *** Waiting for unfinished jobs....
```

### to reproduce
`make ZT_STATIC=1 ZT_DEBUG=0 ZT_SSO_SUPPORTED=0 CC=aarch64-linux-android33-clang CXX=aarch64-linux-android33-clang++ LDFLAGS="-s -L../deps/aarch64" DEFS="-I../deps/include"` ( in ubuntu with ndk toolchain, LDFLAGS and DEFS are set to link the natpmpc library built myself )
or
`make ZT_STATIC=1 ZT_DEBUG=0 ZT_SSO_SUPPORTED=0` ( in termux after installing natpmpc `apt-get install -y natpmpc` )

### possible solution
change that line into

`#if defined(__LINUX__) && ( (!defined(__GLIBC__)) || ((__GLIBC__ >= 2) && (__GLIBC_MINOR__ >= 18)) ) && (!defined(__ANDROID__))`

joseph-henry:
If you're trying to build ZeroTier One for Android there's a lot more that needs to be ripped out and replaced and custom built for the way Android does things. Our Android APK has entirely different packet processing. So this will not work. Going to close this ticket but if you have an issue with the current Android APK please open a ticket for that.

PR #2290, opened 2024-05-10
Update one.cpp

eventlOwOp:
#2289


joseph-henry:
Thanks for this. We'll consider a merge but can you give a little context about what this is fixing for you and how you've tested it?

joseph-henry:
If you're trying to build ZeroTier One for Android there's a lot more that needs to be ripped out and replaced and custom built for the way Android does things. Our Android APK has entirely different packet processing. So this will not work. Going to close this ticket but if you have an issue with the current Android APK please open a ticket for that.

PR #2300, opened 2024-05-21
Brenton/fix detached thread

bostick:
Fix crash in new Android update.

VirtualNetworkFrameFunctionCallback may get called on a thread that is not actually attached to a JVM.

This is a consequence of a change that I actually put in last year:

https://github.com/zerotier/ZeroTierOne/blob/91e7ce87f09ac1cfdeaf6ff22c3cedcd93574c86/node/Switch.cpp#L519

### DIFF
--- a/java/jni/com_zerotierone_sdk_Node.cpp
+++ b/java/jni/com_zerotierone_sdk_Node.cpp
@@ -111,6 +111,44 @@ namespace {
         bool finishInitializing();
     };
 
+    //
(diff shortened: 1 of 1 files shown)


rcoder (approved):
This looks entirely plausible to me as a fix, though I can't claim enough familiarity with our JNI code to understand the implementation in depth. Assuming no one else has concerns and this fixes the immediate crash bug, I'd say (tentatively) :shipit:

Issue #2226, opened 2024-02-14
Possible issue with multipath when used in bridge mode


joseph-henry:
Ticket created for tracking purposes

andrejbinder:
I re-read through all the code and I believe I have found what is causing the problem. When receiving a packet we set the packet flowId to ZT_QOS_NO_FLOW and then only if we hit a packet that is of the VERB_FRAME type we look into its guts and set the flowId based on the actual protocol source/destination port numbers.

The catch is that packets that have been bridged are not VERB_FRAME but they are VERB_EXT_FRAME which contains no code to set the flowId correctly. This is why in-flow assignments do not work for packets from bridged hosts.

Out-flow assignments work perfectly for those because the flowId calculation is done already in Switch::onLocalEthernet of switch.cpp regardless of whether this is a VERB_FRAME or VERB_EXT_FRAME packet.

andrejbinder:
I have also found what looks to me like a typo in the header length discriminator for VERB_FRAME type packets (it was populating flowId only when the packet was big enough to contain a VERB_EXT_FRAME header and at least some payload. I think this was meant to match against VERB_FRAME header size instead. This could make small (VERB_FRAME) packets not to have flowId populated.

andrejbinder:
I put together a small patch to fix both of the issues above:

[ab.patch](https://github.com/zerotier/ZeroTierOne/files/15453381/ab.patch)

Quick testing seems to show that this has fixed the issue and in-flows now work correctly even for bridged packets. Please review the patch and consider upstreaming.

joseph-henry:
This looks good and seems to work. I've made a PR out of it (https://github.com/zerotier/ZeroTierOne/pull/2306) but if you'd rather make your own PR that's fine too. Just let us know what you'd prefer and we can upstream it.

PR #2306, opened 2024-06-07
Fix flow assignment for bridged EXT_FRAME packets

joseph-henry:
This is a PR for @andrejbinder's patch referenced in https://github.com/zerotier/ZeroTierOne/issues/2226

I agree with the rationale and have compiled and tested it and have so far seen no issues.


joseph-henry:
I caught this up with dev and manually verified that the original changes were preserved. I think this is good to go.

laduke (approved):

PR #2312, opened 2024-06-17
Deauth members upon network delete

glimberg:
Add change to explicitly deauthorize members when a network is deleted.  It was possible to get into a state where network members were receiving network status `NOT_FOUND`, yet they could still communicate over the network.


laduke (approved):
Makes sense

PR #2314, opened 2024-06-21
macos: use more portable pwd instead of PWD

Mic92:
In nixpkgs we use coreutils and do not have PWD available. This change have any downsides on normal macOS systems, but helps nixpkgs packaging a bit.

### DIFF
--- a/make-mac.mk
+++ b/make-mac.mk
@@ -1,8 +1,8 @@
 CC=clang
 CXX=clang++
-TOPDIR=$(shell PWD)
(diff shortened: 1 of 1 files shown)


Mic92:
Needed for https://github.com/NixOS/nixpkgs/pull/321509

Mic92:
Friendly ping

joseph-henry (approved):
I agree. Looking at the behavior it seems that `pwd` would be the more portable of the two and I've never seen a Mac without it in recent history.

PR #2330, opened 2024-07-08
break up redis tx inserts into smaller chunks

glimberg:
This is to avoid the following error when the controller loads data:

```
ERROR: Error initializing networks in Redis: ERR Protocol error: -ERR Protocol error: invalid multibulk length
```


laduke:
How does it load eventually load 30001 + items?

glimberg:
It loads all the networks/members from the DB on startup

laduke (approved):

PR #2354, opened 2024-08-29
Multipath improvements

joseph-henry:
A series of small fixes:

 - Fix condition that could prevent `rotate` command from actually rotating links
 - Fix logic bug that could cause a *non*-preferred path from taking the place of a preferred path
 
 More to come.


laduke (approved):
What would be helpful? I don't have the infra to test actual multipathing.

Issue #2361, opened 2024-09-12
Lower ZT_MIN_PHYSMTU


laduke:
ZeroTierOne.h

```
#define ZT_DEFAULT_PHYSMTU 1432

-#define ZT_MIN_PHYSMTU 1400
+#define ZT_MIN_PHYSMTU 576
```
This seems to work, with a local.conf like this:
```js
{ "physical": {"192.168.82.0/24": { "mtu": 789 } } }
```

576 is allegedly the minumum for ipv4. 

If I set MTU to 510, the real packets are 576 in tcpdump, so maybe 510 is the right minimum.

Was looking at this because people on cell connections get stuck sometimes. 
They lower the virtual mtu and that works by side effect. 

This way users can set the MTU based on the physical connection. 

They could use a different value for IPv6 traffic also. 

The constant is used in only 1 place.

pjkundert:
This seems like something that must be auto-configured.  I just had to compute the MTU was 1194 (manually, using `ping -M do -s $(( <MTU> - 28 ))  <host>)` and then set it using `ifconfig <interface> mtu 1194`.

It is unlikely that 100% of ZeroTier users are going to be able to A) determine that this is necessary, and B) figure out how to do it, and B) repeat the process each time every ZeroTier interface is dynamically re-created!

What's the right solution?

destroyer5656:
Yeah I also ran into this issue.

laduke:
This was done  in #2365

PR #2365, opened 2024-09-13
Lower ZT_MIN_PHYSMTU

laduke:
regarding #2361

This constant is only used in one place:

https://github.com/zerotier/ZeroTierOne/blob/d9d58c8bdef89c9beb7325c67777f45c822fc949/node/Topology.hpp#L419-L425

This change can't affect anything unless the user configures this value in their local.conf


joseph-henry (approved):
I haven't personally tested this PR but it does make some sense and I think since it is not going to affect people by default it is safe to merge.

PR #2427, opened 2025-01-27
Fix active backup link selection

william-swarmbotics:
Tested with the "always" link selection policy and a designated primary link.

Old behavior: When the primary link goes down and comes back up, Zerotier does not switch back to it.
New behavior: Zerotier does switch back to the primary link.


Tokisaki-Galaxy:
useful commit!

joseph-henry (approved):

PR #2435, opened 2025-02-10
Support single arch builds on macOS

Enzime:
This is useful for building ZeroTier in Nixpkgs as we only want to build one architecture at once

### DIFF
--- a/make-mac.mk
+++ b/make-mac.mk
@@ -5,7 +5,8 @@ TOPDIR=$(shell pwd)
 INCLUDES=-I$(shell pwd)/rustybits/target -isystem $(TOPDIR)/ext  -I$(TOPDIR)/ext/prometheus-cpp-lite-1.0/core/include -I$(TOPDIR)/ext-prometheus-cpp-lite-1.0/3rdparty/http-client-lite/include -I$(TOPDIR)/ext/prometheus-cpp-lite-1.0/simpleapi/include
 DEFS=
 LIBS=
-ARCH_FLAGS=-arch x86_64 -arch arm64 
+ARCH := x86_64 arm64
+ARCH_FLAGS=$(addprefix -arch ,$(ARCH))
 
 CODESIGN=echo
 PRODUCTSIGN=echo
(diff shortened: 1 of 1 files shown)


make-mac.mk
```diff
@@ -127,12 +129,12 @@ zerotier-one: one
 zeroidc: rustybits/target/libzeroidc.a
 
 rustybits/target/libzeroidc.a:	FORCE
-	cd rustybits && MACOSX_DEPLOYMENT_TARGET=$(MACOS_VERSION_MIN) cargo build -p zeroidc --target=x86_64-apple-darwin $(EXTRA_CARGO_FLAGS)
-	cd rustybits && MACOSX_DEPLOYMENT_TARGET=$(MACOS_VERSION_MIN) cargo build -p zeroidc --target=aarch64-apple-darwin $(EXTRA_CARGO_FLAGS)
-	cd rustybits && lipo -create target/x86_64-apple-darwin/$(RUST_VARIANT)/libzeroidc.a target/aarch64-apple-darwin/$(RUST_VARIANT)/libzeroidc.a -output target/libzeroidc.a
+	[[ "$(ARCH)" == *"x86_64"* ]] && cd rustybits && MACOSX_DEPLOYMENT_TARGET=$(MACOS_VERSION_MIN) cargo build -p zeroidc --target=x86_64-apple-darwin $(EXTRA_CARGO_FLAGS) || true
```
Mic92: I think using `if then` is better because `|| true` would hide build errors

laduke:
I can't test right now but I'll say I sometimes comment out one of the architectures when I'm iterating because it's much faster to compile only one native arch. Would be cool to have if it doesn't break something in our build systems.

make-mac.mk
Enzime: Fixed 👍

Mic92 (approved):

Mic92:
> I can't test right now but I'll say I sometimes comment out one of the architectures when I'm iterating because it's much faster to compile only one native arch. Would be cool to have if it doesn't break something in our build systems.

Should be no change in the behavior, however your use case becomes easier by applying make flags.

Enzime:
@laduke any chance you could take a look again at this PR?

adamierymenko:
Looking at this one before a release -- is there a reason this is important? AFAIK multi-arch macOS binaries only load the target arch and the binary is not large.

laduke:
the OP says something about their downstream package system. 
local compile times being twice as fast when iterating would be good for me.

Enzime:
I've rebased the code on top of the latest changes

The main benefit for this one is being able to optionally build single architecture builds, this is necessary in Nixpkgs as we only support building one architecture at a time.

Nixpkgs is the package set for Nix the package manager which can build and install ZeroTier on both macOS and Linux

barracuda156:
It will be really nice not to hardcode archs, but detect those properly, like it is done on Linux. macOS runs on 5 architectures, not 2.

Issue #2342, opened 2024-08-13
Unable to connect to the ZeroTierOne service HTTP API via IPv4


alexrsagen:
### What you expect to be happening.
Should be able to connect to http://127.0.0.1:9993 (IPv4).

### What is actually happening?
- Unable to connect to http://127.0.0.1:9993 (IPv4)
- Able to connect to http://[::1]:9993 (IPv6)
- Able to connect to http://localhost:9993, as it resolves to http://[::1]:9993 (IPv6)

### Any steps to reproduce the error.
- Install ZeroTier One on Windows 11
- Start ZeroTier One service
- Start ZeroTier UI, observe message "Waiting for ZeroTier system service..." (source: https://github.com/zerotier/DesktopUI/blob/main/src/main.rs#L949)
- `powershell -command "Test-NetConnection -ComputerName 127.0.0.1 -Port 9993"`

### Any relevant console output or screenshots.
```
C:\Windows\System32>whoami
nt-myndighet\system

C:\Windows\System32>netstat -a -o -n | findstr 9993 | findstr LISTENING
  TCP    0.0.0.0:9993           0.0.0.0:0              LISTENING       57760
  TCP    0.0.0.0:9993           0.0.0.0:0              LISTENING       57760
  TCP    [::]:9993              [::]:0                 LISTENING       57760

C:\Windows\System32>netstat -a -o -n | findstr 9993 | findstr UDP
  UDP    192.168.111.146:9993   *:*                                    57760
  UDP    192.168.111.155:9993   *:*                                    57760

C:\Windows\System32>tasklist | findstr 57760
zerotier-one_x64.exe         57760 Services                   0     13 520 K

C:\Windows\System32>powershell -command "Test-NetConnection -ComputerName 127.0.0.1 -Port 9993"
WARNING: TCP connect to (127.0.0.1 : 9993) failed

ComputerName           : 127.0.0.1
RemoteAddress          : 127.0.0.1
RemotePort             : 9993
InterfaceAlias         : Loopback Pseudo-Interface 1
SourceAddress          : 127.0.0.1
PingSucceeded          : True
PingReplyDetails (RTT) : 0 ms
TcpTestSucceeded       : False

C:\Windows\System32>powershell -command "Test-NetConnection -ComputerName '::1' -Port 9993"

ComputerName     : ::1
RemoteAddress    : ::1
RemotePort       : 9993
InterfaceAlias   : Loopback Pseudo-Interface 1
[...]

PR #2438, opened 2025-02-26
Add custom control plane for third-party device vendors

joseph-henry:
This change provides third-party device vendors the ability to manage ZeroTier without exposing the typical HTTPS control plane on `localhost`. It can be enabled at compile-time via `ZT_EXTOSDEP=1`.

There are minor changes to our dependencies which is not ideal but since the changes are so minor and it is so unlikely that they will change again I feel it is acceptable.


ext/prometheus-cpp-lite-1.0/core/include/prometheus/counter.h
```diff
@@ -25,7 +25,18 @@ namespace prometheus {
...
+
+#if UINTPTR_MAX == 0xffFFffFF
+// 32-bit platform
+template <typename Value_ = uint32_t>
+#elif UINTPTR_MAX == 0xffFFffFFffFFffFF
+// 64-bit platform
+template <typename Value_ = uint64_t>
+#else
+#error Unknown platform - does not look either like 32-bit or 64-bit
+#endif
```
glimberg, line 39: So I've seen these changes in a couple of places in the prometheus imports.  Nothing too horrible, for sure, but I wonder if it might be better to have these in a wrapper header just so that we don't have to modify the external stuff.

For instance, we create a `<ztcounter.h>` that includes this block of code, and then #include's the prometheus counter.h.

Just a thought

laduke:
httpcpplib might need updating to fix this issue: 
https://github.com/zerotier/ZeroTierOne/issues/2342

only mentioning it because the change will get reverted if when copy in a new version of the lib

glimberg:
Should probably add an Action that at the very least builds this configuration for PRs just to make sure it's in a state that compiles.

joseph-henry:
> httpcpplib might need updating to fix this issue: #2342
> 
> only mentioning it because the change will get reverted if when copy in a new version of the lib

Noted.

I can come in and fix it when we update the library. I think the vendor in question will be testing this more often and will likely let us know when it breaks.

ext/prometheus-cpp-lite-1.0/core/include/prometheus/counter.h
joseph-henry: Not a bad idea, but it looks like `counter.h` is only ever included from within prometheus so we'd have to patch prometheus to use our wrapper header.

joseph-henry:
@adamierymenko I've changed this from a draft to ready to review. I'll let you take it from here

laduke (approved):

PR #2441, opened 2025-03-03
Run CI on pull requests

laduke:
If the PR is from an external repo,the action won't run without approval.
Right now we can't run actions on external PRs.


glimberg (approved):

laduke:
this might need to get into main before it affects anything actually.

PR #2445, opened 2025-03-26
Allow setting local.conf content from Docker environment variable

glimberg:


joseph-henry (approved):
Looks good, seems like the runners are broken for reasons unrelated to this

glimberg:
yeah.  apt on the validate runner isn't happy for some reason

PR #2474, opened 2025-07-03
Hello encryption

adamierymenko:
Merge hello encryption -- work done a while ago to support encryption of initial HELLO and already supported by our roots.


adamierymenko:
This also contains not-yet-used early scaffolding for a FIPS alt encryption mode for air gapped networks. Not enabled currently so doesn't do anything yet.

laduke:
looks like selftest if failing

laduke (approved):

PR #2482, opened 2025-07-11
Add OpenTelemetry support in Central Controllers

glimberg:
Adds exporting of otel traces exported from the Central build of the controllers.  

Currently only the Controller related parts are wired up for otel, though that could be greatly expanded in the future.  Would also be nice to get the normal builds exporting  otel stats, but that's going to be dependency hell and we'd likely need to update the client build process as a part of it.


aaronjohnson:
I recommend we use https://conan.io/  to manage the upstream source code dependency.

glimberg:
I'm not planning on changing anything about how upstream packages are handled at the moment.  If anything, switching ZT1 to use CMake for builds would make life much easier than conan at this point in time, but I don't think there's an appetite to bite that task off at the moment

adamierymenko (approved):
Looks non-conflicting with other changes, will test more before release.

PR #2485, opened 2025-07-15
Feature: fragmentation metrics

rcoder:
This PR adds packet fragmentation data for both VL1 and VL2 to the Prometheus metrics collected in the core agent.

Credit goes to @aaronjohnson for the bulk of the implementation; I've just tried to rebase, clean up, and test it against the current dev branch.

### DIFF
--- a/node/Metrics.hpp
+++ b/node/Metrics.hpp
@@ -139,6 +139,38 @@ extern prometheus::simpleapi::counter_metric_t db_get_network_list;
 extern prometheus::simpleapi::counter_metric_t db_member_change;
 extern prometheus::simpleapi::counter_metric_t db_network_change;
 
+        // Fragmentation Metrics
+        extern prometheus::simpleapi::counter_family_t packet_fragmentation;
+
+        // VL2 Fragmentation Metrics
+        extern prometheus::simpleapi::counter_metric_t vl2_oversized_frame_tx;
+        extern prometheus::simpleapi::counter_metric_t vl2_would_fragment_or_drop_rx;
+
+        // VL1 Fragmentation Metrics
+        extern prometheus::simpleapi::counter_metric_t vl1_fragmented_tx;
+        extern prometheus::simpleapi::counter_metric_t vl1_fragment_rx;
+        extern prometheus::simpleapi::counter_metric_t vl1_reassembly_failed_rx;
+        extern prometheus::simpleapi::counter_metric_t vl1_fragment_without_head_rx;
+        extern prometheus::simpleapi::counter_metric_t vl1_fragment_before_head_rx;
+        extern prometheus::simpleapi::counter_metric_t vl1_duplicate_fragment_rx;
+        extern prometheus::simpleapi::counter_metric_t vl1_duplicate_head_rx;
+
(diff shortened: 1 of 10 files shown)


node/Metrics.hpp
```diff
@@ -139,6 +139,38 @@ extern prometheus::simpleapi::counter_metric_t db_get_network_list;
 extern prometheus::simpleapi::counter_metric_t db_member_change;
 extern prometheus::simpleapi::counter_metric_t db_network_change;
 
+        // Fragmentation Metrics
```
laduke, line 142: very minor comment but the indenting is different here than above. 
maybe we need to add clang format to ci

node/Switch.cpp
```diff
@@ -1022,6 +1042,9 @@ unsigned long Switch::doTimerTasks(void* tPtr, int64_t now)
 		if ((rq->timestamp) && (rq->complete)) {
 			if ((rq->frag0.tryDecode(RR, tPtr, rq->flowId)) || ((now - rq->timestamp) > ZT_RECEIVE_QUEUE_TIMEOUT)) {
 				rq->timestamp = 0;
+				if ((now - rq->timestamp) > ZT_RECEIVE_QUEUE_TIMEOUT) {
```
laduke, line 1050: Looks like `rq->timestamp` always 0 at this point?

node/Metrics.hpp
laduke: `make clang-format` exists as of a recent commits

laduke:
Any info on how to verify or test some of these metrics? 

If I do `ping -s 9000 10.2.0.2` should I expect something to change in metrics.prom? stuff like that

rcoder:
> Any info on how to verify or test some of these metrics?
> 
> If I do `ping -s 9000 10.2.0.2` should I expect something to change in metrics.prom? stuff like that

If you're running this version, the fragmentation metrics should indeed appear in `metrics.prom`; figuring out how to trigger actual fragmented traffic is left as an exercise for the reader. ;)

joseph-henry:
I'm concerned about an `.Observe()` call being in the `put()` function in the tap. Has anyone tested the performance impact of this?

aaronjohnson:
My performance measurements to date have not indicated a performance impact.    That said, I have not re-created non-zero counters.   I'd like to confirm this is true for customers as well.

@joseph-henry would the counter need to increment? Or just that the code path is followed.

I have not been able to set up the scenario where these code branches are reached using netns.

This is in support of collecting detailed information about a theory of VL1 MTU << VL2 MTU implications from impacted customers.

Is there another more performant way to collect this than using prometheus?

Would a local.conf non-default switch to enable these counters suffice?

aaronjohnson:
> Any info on how to verify or test some of these metrics?
> 
> If I do `ping -s 9000 10.2.0.2` should I expect something to change in metrics.prom? stuff like that

Also the `-v` option.

aaronjohnson:
Performance comparison proposal.

https://github.com/zerotier/network_test_lab/compare/product-benchmark-comparison-fragmentation-metrics?expand=1

joseph-henry:
This is still in my queue, not forgotten

PR #2489, opened 2025-07-19
Fix link select field in bond CLI

william-swarmbotics:
This makes the active backup link select parameter show up in `zerotier-cli bond ___ show`. Previously it always showed 0.

```
Bond                   : active-backup
Link Select Method     : 3
```


william-swarmbotics:
Related: https://github.com/zerotier/docs/pull/263

PR #2494, opened 2025-08-03
Add support for automatic Firewall Cloud Service Provider Rule Set management (UDP port) and flexible listening TCP port in ZeroTier TCP Proxy

lferrarotti74:
This PR enhances the ZeroTier TCP Proxy by adding support for configuring different internal and external TCP ports. This feature allows the proxy to listen on one port internally while exposing a different port in cloud firewall rules.

### Changes Made

1.
   Configuration Updates :
   
   - Added new externalTcpPort configuration option in local.conf and cloud configuration files
   - Maintained backward compatibility with existing configurations that only specify tcpPort
   - Updated configuration examples and documentation

2.
   Core Implementation :
   
   - Modified FirewallManagerFactory to support both internal and external TCP ports
   - Updated LinodeFirewallManager to handle different internal/external port configurations
   - Added logic in tcp-proxy.cpp to load and manage both port settings

3.
   Firewall Rule Management :
   
   - Enhanced firewall rule descriptions to indicate when internal/external ports differ
   - Updated rule synchronization logic to use the external port for firewall rules

4.
   Documentation :
   
   - Updated README_CLOUD_INTEGRATION.md with information about the new feature
   - Added examples showing how to configure different internal/external ports

### New Capabilities
This change enables several new use cases:

1.
   Port Standardization : Organizations can expose standard ports (like 443) externally while using different internal ports to avoid conflicts with other services.

2.
   Security Enhancement : The ability to use non-standard internal ports while exposing standard ports externally can add a layer of security through obscurity.

3.
   Flexible Deployment : Environments with port restrictions can now run the TCP proxy on available internal ports while maintaining expected external ports.

4.
   Cloud Provider Compatibility : Some cloud providers have restrictions on certain ports - this feature allows working around those limitations.

### Backward Compatibility

This implementation maintains full backward compatibility with existing configurations. If only tcpPort is specified, the external port will default to the same value as the internal port, preserving the current behavior.

### Testing

The changes have been tested with:

- Configurations specifying only internal port
- Configurations with different internal and external ports
- Linode firewall rule creation and synchronization

### Future Work
Potential future enhancements could include:

TBD

### DIFF
--- a/tcp-proxy/Makefile
+++ b/tcp-proxy/Makefile
@@ -2,8 +2,27 @@ CXX=$(shell which clang++ g++ c++ 2>/dev/null | head -n 1)
 
 INCLUDES?=-I../ext/prometheus-cpp-lite-1.0/core/include  -I../ext/prometheus-cpp-lite-1.0/simpleapi/include
 
+# Add libcurl dependency for provider targets
+LIBS=-lcurl
+
+# Default target - original version without cloud provider functionality
 all:
 	$(CXX) -O3 -fno-rtti $(INCLUDES) -std=c++11 -pthread -frtti  -o tcp-proxy tcp-proxy.cpp ../node/Metrics.cpp
 
+# Provider target - includes new functionality
+provider: tcp-proxy-enhanced
+
+tcp-proxy-enhanced:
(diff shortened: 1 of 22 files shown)


lferrarotti74:
It was an error, PR requested too early in action.
Still multiple tests are needed before pushing this PR.
Sorry ;)

PR #2511, opened 2025-09-11
Shorten/clean README.md

Fightersbane:


laduke:
I think the build and platform notes are useful internally and not written down anywhere else. I don't keep how to build on freebsd for example in my memory. 

I would not be sad about the layout section not being there. 

As a developer when I see I repo readme I just want to know how to build and run the program. Whether that perspective applies to this repo I don't know one way or the other.

Fightersbane:
That's a fair point and the build/run sections aren't particularly long anyway. Doesn't make a lot of sense to keep those on the documentation site as opposed to on the repo.

joseph-henry:
I like where this is going, just a few things:

- Title is duplicated
- links to architecture doc and controller are broken, (controller was moved to `nonfree/controller`

(opinions)

- I think the build instructions should be linked to at the top with something like `How to build`
- I think the "project structure" blurb could actually be split into a few tiny README.md's in each directory. Even if it's just a sentence or two for each. I usually click around a little before I read and if I see a README with context in a folder it really helps. But again this is just my preference.

I did not yet check the `build.md` for accuracy but if you didn't change the contents then I'd say we're close to being able to merge this.

Fightersbane:
build.md is just build from the existing info from the old readme

Fixed errors, actually agree with you on having many tiny little README's but for the purposes of this PR I just want to get this out

rcoder (approved):
:shipit:

PR #2535, opened 2025-11-03
Ensure members in PacketMultiplexer are initialized

cspiegel:
_enabled is set to true by setUpPostDecodeReceiveThreads(), so disabled until then, but the constructor wasn't initializing it. _concurrency is not being used before being set but for safety's sake, ensure it has a starting value as well.

Also, remove the vestigial _rxThreadCount, which is no longer used.


joseph-henry (approved):
Thanks for taking care of this.

Issue #2492, opened 2025-07-30
allowDNS doesn't work on Linux: `LinuxEthernetTap::setDns` doesn't do anything


jfly:
I'm trying to enable `allowDNS` on Ubuntu, and it's not doing anything. No changes to `/etc/resolv.conf`, nothing.

I dug into the code, and it looks like this is because `LinuxEthernetTap::setDns` simply is a no-op:

https://github.com/zerotier/ZeroTierOne/blob/185a3a2c76e6bf1b1c0415871f43076638eb007c/osdep/LinuxEthernetTap.hpp#L61

It found some threads online talking about this:

- https://www.reddit.com/r/zerotier/comments/lmei2w/how_do_you_set_up_the_new_dns_feature/
- https://discuss.zerotier.com/t/dns-servers-ignored-on-ubuntu/25231

But I couldn't find any official docs talking about this, so I thought I'd ask here:

1. Is this feature WIP, or intentionally not supported on Linux?
   - I understand that there are lots of ways to do networking and DNS on Linux. Supporting systemd-resolved would work well for a lot of Linux users (myself include). See [How to Implement this in Your VPN Software > If managing a corporate VPN](https://systemd.io/RESOLVED-VPNS/#how-to-implement-this-in-your-vpn-software) for how to configure dns for a specific tld, which would work well with ZT's model.
3. Would you accept a PR adding some logging to `LinuxEthernetTap::setDns` with some explanation about the state of this? The current behavior is very confusing.

laduke:
Thanks for writing. It's definitely not implemented. If you didn't see any notes about it, we need to improve the docs. 
There are so many ways to do dns on linux, we didn't implement any of them. There's a systemd based tool here 
https://github.com/zerotier/zerotier-systemd-manager

jfly:
Thanks for the pointer to https://github.com/zerotier/zerotier-systemd-manager, I'll check it out.

I do believe a warning would be a godsend for future users. How does https://github.com/zerotier/ZeroTierOne/pull/2493 look?

jmariondev:
I've put up a PR ( #2575 ) with an implementation to call `resolvectl` as necessary, though I've only tested this on Fedora.

This required a Polkit rule to allow the `zerotier-one` user to manipulate `systemd-resolved` but this may need to be generalized more to work on a wider range of systems.

PR #2575, opened 2026-04-05
WIP: Implement Linux DNS support via systemd-resolved

jmariondev:
Implements `LinuxEthernetTap::setDns`. The `allowDNS` network flag now works on Linux, matching existing Windows and macOS behavior.

This partially addresses #2492 , though has only been tested on Fedora 43. The polkit approach may need to be fixed for Debian/Ubuntu.

# Approach

- New `LinuxDNSHelper` class following the same pattern as `WinDNSHelper` and `MacDNSHelper`
- Calls `resolvectl dns <iface> <servers>` and `resolvectl domain <iface> ~<domain>` to configure per-interface DNS routing via `systemd-resolved`, the recommended approach for VPN-like services
- The `~` domain prefix configures a *route-only* domain (not a search domain), so queries matching that suffix are routed to the ZT DNS servers without polluting global resolution
- Cleanup via `resolvectl revert <iface>` in the destructor and when `allowDNS` is toggled off
- Subprocess spawning uses `fork`/`exec` matching the existing `_routeCmd` pattern in `ManagedRoute.cpp`

## Non-systemd systems

Checks for `/run/systemd/resolve/stub-resolv.conf` before attempting any `resolvectl` calls. If `systemd-resolved` is not active, `setDNS` logs a warning pointing to #2492 and returns.

## polkit rule

**(This was developed for Fedora, I'm not sure of the Debian/Ubuntu situation here)**

ZeroTier drops root privileges to the `zerotier-one` user at startup (retaining `CAP_NET_ADMIN`/`CAP_NET_RAW`). However, `systemd-resolved` checks polkit identity — not capabilities — for DNS changes. A polkit rule is included and installed by the RPM spec to grant the `zerotier-one` user access to `org.freedesktop.resolve1.{set-dns-servers,set-domains,revert}`.

## Interface name vs network ID

Unlike Windows/macOS helpers which take a network ID, `LinuxDNSHelper` takes an interface name because `resolvectl` operates on interfaces.

----

*This PR was created with help from Claude.*

### DIFF
--- a/ext/installfiles/linux/zerotier-one-dns.rules
+++ b/ext/installfiles/linux/zerotier-one-dns.rules
@@ -1,0 +1,13 @@
+// Allow the zerotier-one service user to configure per-interface DNS
+// via systemd-resolved. Required because zerotier-one drops root
+// privileges to the zerotier-one user while retaining network
+// capabilities, but systemd-resolved checks polkit (not capabilities)
+// for DNS configuration changes.
+polkit.addRule(function(action, subject) {
+    if ((action.id == "org.freedesktop.resolve1.set-dns-servers" ||
+         action.id == "org.freedesktop.resolve1.set-domains" ||
+         action.id == "org.freedesktop.resolve1.revert") &&
+        subject.user == "zerotier-one") {
+        return polkit.Result.YES;
+    }
+});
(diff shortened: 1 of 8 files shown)


jmariondev:
I've added a second commit that focuses on subprocess security.

jmariondev:
Rebased over current `dev`.

I have been running this on my Fedora machine since submitting this and it's worked flawlessly.

I could use help from someone familiar with Debian-family distros to make this work there too :pray:

